PR #21, opened 2015-07-29
License moved to root directory

paulvollmer:
because at the root directory is no license file...


el-j:
seems to make sens. thanks

PR #23, opened 2015-07-30
DQuid IO GPRS first release

vitopoleo:


el-j:
some bugs in this commit: 
1. breadboardimage can not befound 
2. no buses between same connections (e.g.)
if you need any help please let me know

PR #25, opened 2015-09-04
4060 - 14-stage ripple-carry binary counter/divider and oscillator

neutmute:
Hi - this is my first Fritzing part. 

It is a [4060 14-stage ripple-carry binary counter/divider and oscillator](http://www.nxp.com/documents/data_sheet/HEF4060B.pdf).

The [google code parts issue](https://code.google.com/p/fritzing/issues/detail?id=2753) is now read only so I figure this repo is how to contribute parts?

Notes
* I've read the contribution guidelines but it doesn't indicate whether I should putting parts in core or `contrib`. I chose `contrib`.
* To get the base line part, I copied from a `ULN2003A` which you can see references to in the source - does that need cleaning up?
* I assume the generic 4060 name is ok and it doesn't need to be more specific?
* The exported part had GUID's in the filenames when unzipped - which I cleaned up, I'm assuming I got the file names correct. Is there a way to compile the part again to ensure I got this correct?

any problems, let me know and I will correct
cheers!


el-j:
hej, thanks for the part. 2 things:
-> could you please correct the svg-name in the parts-fzp
-> the reference-file should be the 4060 fzp

neutmute:
done

PR #26, opened 2015-09-14
LittleBits w29 Perf

shaiss:
Added the LittleBits w29 Perf module


el-j:
thanks for contributing this part. sadly it is not possible to the fzpz file format. it is better to put the raw files in the folder structure:

fzp -> core-folder
breadboard svg -> svg/breadboard/
schematic svg -> svg/schematic/
pcb svg -> svg/pcb/

you can use a python script to clean up the fzpz easily:
https://github.com/fritzing/fritzing-parts/blob/master/part-gen-scripts/misc_scripts/fzpzclean.py

thanks for your understanding
greetings

shaiss:
Could I just export as svg from fritzing directly?   The python script looks good, but I'm on win and would have to install python just to run that.

shaiss:
Could I just export as svg from fritzing directly?   The python script
looks good, but I'm on win and would have to install python just to run
that.

PR #46, opened 2016-04-23
add new custom part tilt switch sw-200d

sayanee:
I have added a [note in the Fritzing forum too](http://forum.fritzing.org/t/sw-200d-tilt-switch-new-custom-part/1073)

**Manufacturer and name:** YUSAN, tilt switch SW-200D

**Original photo**

sw-200d

**Working example with a breadboard circuit**

image

**Breadboard photo**

sw-200d-breadboard

**Icon pic**

sw-200d-icon

**PCB pic**

sw-200d-pcb

**Schematic pic**:

sw-200d-schematic

### DIFF
--- a/core/SMD_LGA-12.fzp
+++ b/core/SMD_LGA-12.fzp
@@ -25,7 +25,7 @@
 </layers>
 </breadboardView>
 <schematicView>
-<layers image="schematic/sparkfun-sensors_bma180_schematic.svg">
+<layers image="schematic/IC12.svg">
 <layer layerId="schematic"/>
 </layers>
 </schematicView>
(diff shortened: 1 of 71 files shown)


el-j:
thanks for the part. it is added to the core library now. 

i have to do some fixing because of the file names. if a fzpz is extracted it has the definitions
for the folderstructure in the filename: 
part.icon.xxxxxxxxxxxx.svg

that confuses fritzing, so you have to do some file name cleaning or use the python script from:
https://github.com/fritzing/fritzing-parts/blob/master/scripts/fzpzclean.py

nice work. thanks again

PR #47, opened 2016-05-16
Adding part for 96boards Consumer Edition

rsalveti:
96boards is an open hardware specification compatible with several
boards (e.g. Dragonboard410c, HiKey). This part is to cover the boards
compatible with the Consumer Edition specification.

Signed-off-by: Ricardo Salveti rsalveti@rsalveti.net


el-j:
thanks for the new part. sadly the part will not work properly. 
-> in the pcb-svg is just one copper-layer defined. that will cause a short in pcb-view and gerber export.
-> if you look into the svg-code there are many translate and transform parameters, as well as many unnecessary empty groups. these parameters can have an really ugly output in the geber export

PR #51, opened 2016-06-07
Dagu DGServo 9g

sgparry:
Dagu Servo Parts and Dagu bin. Further Dagu robot parts to follow.
breadboard2
pcb
schematic_view2


el-j:
hello stephen, nice work so far. 
why do you have the "non-pcb-view" variants? this causes rats-nest/air-wires to disappoint in pcbview in some cases. and i think we do not need the headers because they are normal generic 3pin headers.

could you sort them out?

sgparry:
el-j: Thanks for the feedback.

The non-pcb variants are there because the normal use case is that the servo is connected via a supplied pre-attached female connector to either header pins on an Arduino shield or male header pins on a PCB, breadboard or stripboard. The necessary copper holes are therefore already either on the male header pins or the shield. Any such holes on the motor or female connector part are therefore a nuisance. It also means you can label the motor in schematic view as a motor, not as a jumper.

The two reasons I have the custom 3pin headers are a) to avoid having such holes on the female part b) to allow all the pins on both connectors to be correctly labelled in schematic view. With non-keyed connectors it is all too easy to end up with plugs that are connected backwards, c) the real-life female connector actually comes attached as part of the servo; if there was a way of attaching it on the ends of the bendable wires as a sub-part in Fritzing I would do that.

With regard to the air-wires, please can you screenshot me an example and show me exactly what you think should not be happening. I might be able to find a better way of doing things.

sgparry:
OK I have done more research into how Fritzing handles 'off-board' parts - I can see now why the parts need PCB holes even if they are off-board, ie. to help with the routing.
I will also need to rethink how I do the through-board holes; they should be a sub-part in PCB view, but PCB view doesn't handle sub-parts. Any thoughts?

sgparry:
Can you review my latest changes for approval please?

el-j:
there is one breadboard image missing:
image

in one fzp the filename is missing a "_breadboard.svg".
it's the  dagu pan and tilt fzp. 

after that change i can merge your branch. thank you for all the good work

sgparry:
Thanks for spotting that. I have corrected. This is just the beginning...

sgparry:
I am working on some more parts that I'd like to include in Dagu bin; Could you please check and commit this request so I can base the new one on this? Thanks.

PR #56, opened 2016-06-17
Added JST-VH 3 pin male header and female connector types

sgparry:
These connectors are commonly used as connectors for Dagu robot main
motors. They do not play nicely with 0.1 in strip or breadboard. The male
can be mounted on stripboard if the centre hole is used for location purposes,
holes drilled between the tracks and solder traces added to make high
current double tracks for the outer connectors.

The Fritzing part in this case has pin0 as the centre pin so that the male
header locates symmetrically on stripboard. The labelling is still pin 1
is the leftmost however.

jst-vh-breadboard
jst-vh-pcb
jst-vh-schermatic

### DIFF
--- a/bins/core.fzb
+++ b/bins/core.fzb
@@ -899,6 +899,22 @@
         </views>
       </instance>
   
+      <instance moduleIdRef="JST-B3P-VH" modelIndex="46" path="JST-B3P-VH.fzp">
+        <views>
+          <iconView layer="icon">
+            <geometry z="-1" x="-1" y="-1"></geometry>
+          </iconView>
+        </views>
+      </instance>
+  
+      <instance moduleIdRef="JST-VHR-3N" modelIndex="46" path="JST-VHR-3N.fzp">
+        <views>
+          <iconView layer="icon">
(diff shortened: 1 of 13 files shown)


sgparry:
I know we've all made a lot of pull requests lately, but if someone could please review this one, then I can hopefully move on to some parts that depend on it. Thanks.

el-j:
thanks for the parts.
so there is a problem with the female header. in pcb view when you try to connect to one of the connector-pins1-3 (ids: 0-2) it randomly chooses pin4-6 (ids: 3-5) sometimes. 

i think you have to move the pins to a group outside the copper groups with and id "unknown"
in the svg and switch the layer in the fzp to unknown for connector id 3-5 for the pcb connectors.

sgparry:
I've done as you suggested. Can you give them a try please? I only managed to reproduce the problem in one way, which seems fixed now.

sgparry:
Please can you review the changes? Thanks.

el-j:
sorry again, good work! is merged now

PR #60, opened 2016-06-20
Added parts for the 6 40xx series standard quad x two input logic gates.

sgparry:
These parts include symbols in schematic view with sub-parts allowing you
to locate each gate device individually.

40xx_schematic


el-j:
amazing work! thank you!

one little thing. the right bottom corner gate is 0.05inch off the grid. 
image

it seems that it uses the singel connector to define it's position. all the others use the side with the 2 connectors, so they fit perfectly. maybe you can change that before the merge?

sgparry:
Weird! I cannot reproduce that on my system at all. The gates go on perfectly every time. They snap to grid on the rear legs of the part. If you try moving the bottom right part separately, does it snap on the front or the rear legs?

sgparry:
Extensive rework because of issues with subscripts and with internally connected sub-parts (bug).
New parts now exist as two variations - a single part version with normal pin order and simpler multi-part version that follows logical rather physical pin order. Removed subscripts until Fz can handle them.

40xx_schematic_v2

sgparry:
As can be seen from the git log, reworked the components for strictly 0.1in grid. Also reworked all the parts to make them more consistent in size and layout. Separated single and multi-part into different families.
40xx_schematic_v3
two-bit-adder-v2

sgparry:
Hi el-j,
Really sorry to pester like this but these and #56 are (as far as I can tell) now ready for use. I am getting requests for these and I need #56 for further parts. Please can you give them a final check and merge them?
Thanks
Stephen

sgparry:
what am I doing wrong please?

el-j:
sorry, for the delay! special and great work! merged!

sgparry:
Thanks @el-j, that is very much appreciated.

PR #62, opened 2016-06-21
Add WeMos bin

mcauser:


mcauser:
WIP - do not merge.
The WeMos bin should only have 1x part, with 4x variants:

* female headers above
* male headers above
* male headers below
* no headers

When someone added the "_fixed" version, it looks like 2 of them were dropped.

el-j:
hello mcauser, thanks for all your work. 
please understand that we do not want to make a single bin for every part. 
we do not want to have millions of bin symbols. our decision goes to a minimum
of 12 parts to make a new bin. i hope you understand that position. 
anyway, nice parts can be merged to the core parts and can maybe take place in the core
bin file. shorted in the correct section

thanks

PR #63, opened 2016-06-21
Add MicroPython bin

mcauser:
Adds the MicroPython pyboard v1.0 to a new bin.
I'm working on another 3 MicroPython boards which will also be added here.

The original pyboard: https://github.com/mcauser/Fritzing-Part-Micro-Python-pyboard


el-j:
see conversation here: https://github.com/fritzing/fritzing-parts/pull/62

mcauser:
No problem. Group them in the core bin under microcontrollers?

el-j:
yeah maybe you can make one change to the core.bin in another pull request? that would be really nice. maybe there can be a new group like: wlan devices ... e.g. ... you can be creative on that. the core library can get more sorting at all. 

mainly the aim behind the core bin is that it shows the most relevant part of a good organised family. 
the question behind all that: what is the right family. :D
if you look at the arduino boards, they are all the same family, so you can choose from what you want. 
in case of wlan-devices it is maybe more suitable to have a "wlan-microcontroller" section and the families are maybe "connected device" ... i like the idea that parts are not arranged by manufacturer...

PR #65, opened 2016-06-22
Added Camdenboss 90 degree 0.2in pitch screw terminals, 2 to 12 pins.

sgparry:
camdenboss-ct0158

### DIFF
--- a/bins/core.fzb
+++ b/bins/core.fzb
@@ -907,6 +907,14 @@
         </views>
       </instance>
   
+      <instance moduleIdRef="Camdenboss_CTB0158-4_5_08mm_pitch_90deg_terminals" modelIndex="46">
+        <views>
+          <iconView layer="icon">
+            <geometry z="-1" x="-1" y="-1"></geometry>
+          </iconView>
+        </views>
+      </instance>
+  
       <instance moduleIdRef="DB25F" modelIndex="46" path="DB25F.fzp">
         <views>
           <iconView layer="icon">
--- a/core/Camdenboss_CTB0158-10_5_08mm_pitch_90deg_terminals.fzp
+++ b/core/Camdenboss_CTB0158-10_5_08mm_pitch_90deg_terminals.fzp
@@ -1,0 +1,358 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<module fritzingVersion="0.2.2.b.03.04.2550" referenceFile="screw_terminal_10_200mil.fzp" moduleId="Camdenboss_CTB0158-10_5_08mm_pitch_90deg_terminals">
+ <version>4</version>
+ <author>Stephen Parry</author>
+ <title>Camdenboss CTB0158-10</title>
+ <label>J</label>
+ <date>Mon Jun 20 2016</date>
+ <tags>fritzing core</tags>
+ <properties>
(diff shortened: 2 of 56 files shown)


el-j:
thanks for the parts. they look really nice.
i take a review and there is something going wrong in the pcb view.
the coppertraces connecting to some place sometimes. i assume it happens
because there are more connectors defined as used. 
image

el-j:
ok i see the problem. you do not need to define the terminals as connectors. the terminal point is just for the schematic view to set the connection point for a wire. it is simply a rectangle without filling or any "size" at the end of the connector-path. if you look at a normal generic header part, you find a good example.
image

sgparry:
I think actually it might be the same problem as with the JST part? What I have seen on the forums suggests the size of the connector_N_terminal shape is irrelevant, it just uses the centre point. I gave mine a size because zero sized shapes don't show in inkscape. I will try both fixes and see if they help.

sgparry:
OK - can you try those please?
Thanks very much BTW for all the careful quality control work you do - it is amazing.

el-j:
hello again, i checked your last changes.
now it seems ok, but i think it is confusing having the pcb-connectors and the outgoing connectors in one part. it's quiet a bit strange to use. do you think it will not confuse people? 

image

image

image

sgparry:
I am open to alternative suggestions; can you sketch one? The breadboard 
and pcb views are dimensionally and visually accurate - they should make 
sense to anyone who has the actual part. The schematic view I am not 
sure of - I have not been able to find an IEEE symbol for this kind of 
connector nor an example of how to use it - I just adapted the symbol 
from the blue screw terminals in Fritzing. I am fairly sure however that 
it should look something like this when used:


Or possibly like this:


I am not sure which is IEEE correct. Neither is possible without the 
separate connectors.
Please let me know what you think.


On 5 July 2016 18:55:38 BST, el-j <notifications@github.com> wrote:

    hello again, i checked your last changes.
    now it seems ok, but i think it is confusing having the
    pcb-connectors and the outgoing connectors in one part. it's quiet a
    bit strange to use. do you think it will not confuse people?

    image
    <https://cloud.githubusercontent.com/assets/2795534/16594278/f98e40be-42e9-11e6-9979-362918cc6410.png>

    image
    <https://cloud.githubusercontent.com/assets/2795534/16594269/ee23620e-42e9-11e6-8236-e808fb7fc4ea.png>

    image
    <https://cloud.githubusercontent.com/assets/2795534/16594304/14848c52-42ea-11e6-9e76-d539d53c8f14.png>

    —
    You are receiving this because you authored the thread.
    Reply to this email directly, view it on GitHub
    <https://github.com/fritzing/fritzing-parts/pull/65#issuecomment-230552854>,
    or mute the thread
    <https://github.com/notifications/unsubscribe/AER9PVH8aegvlagM20QrDfC1Pd7SOLcIks5qSpqagaJpZM4I7V9P>.

sgparry:
For some reason the images didn't work in an e-mail reply, even after I added them manually, so here they are:
ctbtest
ctbtest2

el-j:
ok that makes sens. lets see how others react on the part and this behavior. thank you very much.
merged and closed.

PR #67, opened 2016-06-23
Sharp PC817 PC847x

ThomDietrich:
Adds parts PC817 and PC847X, two popular simple optocouplers. I made sure they are in the same family as the two existing optocouplers and tested everything.


el-j:
hello nice work so far! thank you.

i reviewd your request and one of your pcb-svg has just one copper layer defined.
the PC847X pcb svg.
this will cause a short in production. please add the other copper1 group. 
another thing is the "translate" in the copper group. this can cause confusion
during the gerber export. maybe you get ride of this as well.

ThomDietrich:
Hey el-j!
I see that you fixed the problem https://github.com/fritzing/fritzing-parts/commit/7eabd2e8b091756aff12a3a835babc2a0d13ac5c and merged everything. Thanks for that!

This was my first fritzing part, so could you elaborate a bit? Reading through [this](https://github.com/fritzing/fritzing-app/wiki/2.1-Part-file-format#through-hole-vs-smd-parts) I get your changes. Why would not having "copper1" create a short? Would it not just not create copper on only one side of the pcb? Also, how could I have created that copper1 group in Inkscape / a better software?

Is getting rid of translate still needed? Again, how would I do that in Inkscape?

el-j:
the missing copper layer causes the short because the holes are connected with the bottom side "through the hole". so that means, the copper-donut on top has copper as well in the "hole". if you just define one copper layer, and make a copper fill, you will see that in gerber, the holes connecting all together on the side where the part has the missing copper side.

i don't know much about inkscape. if it is not possible, you can simply add it via a text editor (like i fixed you svg):
<g id="copper0">
**<g id="copper1">**
the connectors
**</g>**
</g>

untested: (i am using mostly Ai myself)
how to clean svg's and setup inkscape to delete transforms, and make real positioning of objekts:
http://graphicdesign.stackexchange.com/questions/18088/exporting-an-object-as-svg-from-inkscape
http://stackoverflow.com/questions/13329125/removing-transforms-in-svg-files

a webbased svg editor. it seems to remove really much everything that is not needed for a svg.
but maybe it removes to much. has to be tested with fritzing:
http://editor.method.ac/

ThomDietrich:
What a mess. The links were not helpful and Inkscape kept injecting transformations - even worse, it started to restructure everything and delete ids like copper1.
In the end I changed everything in a text editor. Hope everythings fine now.

PR #117, opened 2018-01-04
Update the Raspberry PI Zero part with many changes

vanepp:
With the warning I'm just learning to use github (within the last week), I think this should be a pull of my modified Raspberry PI Zero part. It works fine in Fritzing, but is a substantial change (including all the pin numbering from the original part being different to better match the PI documentation). It makes it through my part checking script without complaint as well. It looks to me like the problem on your last commit is that the fzp file got completely deleted in the push if you would rather try and fix that up rather than merge this (although I'd like this version to go in at some point, and then I'll modify all the other versions to match this one as well so they are all consistent).

Peter


aknoerig:
Awesome, thanks! @el-j can you do this by deprecating the old part?

el-j:
done. thank you really much

PR #119, opened 2018-01-08
IchigoJamT(Jan 8th,2018)

T-Sa-sa-ki:
I reexamined it so as not to conflict with other parts information.


el-j:
merge to nightlyparts, QA done, pushed to master. 
thank you for the new part.

PR #120, opened 2018-01-13
Update the RioRand_LM2596 part in core

vanepp:
This new part should fix issue 74 (which has a lot of problems, 2 pins don't work in breadboard as they are bendable legs but not outside the view box etc). I did a version of this in the forums about the time this request was made, but this time I used the dimensions in the drawing in issue 74 for both breadboard and pcb. It has been post processed by the part checking script. I exported the original part and made the changes to that, is it easier for you if I make the changes against the original part rather than do this?

Peter


el-j:
merge to nightlyparts and pushed to master afterwards
thanks you for the fix

PR #122, opened 2018-01-28
Update CONTRIBUTING.md

CWempe:
I think the paths in the CONTRIBUTING.md were wrong and I fixed it.

I also added some info about where the user can find the separate files of custom parts.

Please check if I am correct. 😄


CWempe:
The commit about the GoldCap should not be in here.

Will try again.

PR #128, opened 2018-02-02
merge master runs travis

el-j:


KjellMorgenstern:
It think this PR is outdated. Can it be closed?

ovidiub13:
I agree, and we could probably also remove the `test` branch as it is not a part of the **Git Flow** process.

PR #130, opened 2018-02-06
Add part Proto_shield_Mega_REV3

Pikrass:
The proto shield for Arduino Mega ([link to Arduino store](https://www.arduino.cc/en/Main/ArduinoMegaProtoShield)). Based off the "normal" proto shield.


el-j:
really nice work! love it. 
but as far is can see, there is not enough space between the edges of the connectors and the copper-connection. 
image

the other thing is the drc.
image

all the connection-path between the in and out connectors have a problem. maybe you could try to split the connection-path in two paths that overlap each other... i am not sure how to solve this problem. did you see any other parts that have a copper-connection-path between connectors without drc errors?

Pikrass:
Yeah, I wasn't too fond of what I did there, I should rework it when I get the time. If you have any idea on how that could look nicer...
I'm not too familiar with the DRC so I'm not sure how to solve this either. I don't think I've encountered a part with a similar layout.

el-j:
i guess if you remove the rectangle shape from all the connectors the drc should be fine.
be careful keeping the connector id's from the svg!

PR #136, opened 2018-03-08
Added missing whitespace between attributes

paulvollmer:
some linter will throw an error if there is no whitespace


paulvollmer:
@el-j ^^ some javascript parser will `throw an error`

PR #139, opened 2018-03-13
Seeedstudio LinkIt One board

KingDarBoja:
Added an improved version of Seeedstudio LinkIt ONE board. Since I couldn't found the original one at mediakit labs, I made my own version.

[LinkIt One Fritzing discussion](http://forum.fritzing.org/t/linkit-one-development-board/5586/1)

**NOTE:** _PCB layout is based on Arduino MEGA from fritzing while breadboard got made in order to make all connector pin fit into 0.1 in Fritzing grid._


el-j:
thank you, pushed to nightlyparts

PR #146, opened 2018-03-24
Diac diode db3

KingDarBoja:
Added DIAC diode db3 as requested at [Fritzing discussion: Where the DIAC on Fritzing](http://forum.fritzing.org/t/where-the-diac-on-fritzing/5678).

EDIT#1: Noticed I did the checkout from master to the _new part branch_ so that's why there are 3 extra commits.

EDIT#2: I reverted those commits to clean up the pull up request.


el-j:
nice work. merged

PR #170, opened 2019-02-01
New part Onoffshim

bortek:
Adding new part OnOffShim for RPi after following instructions which were actually quite good.

Spend though two days drawing this part mostly having problems with Inkscape. Did some manual adjustments in XML files too. 

Tested drawing and verifying PCB/schematic etc and all looks good to me.

Please take a look and approve.


bortek:
Is someone looking at the PRs ?

el-j:
seems good. thanks. nice work

bortek:
Yeyii. Thanks for approving.

PR #177, opened 2019-03-09
Nightly parts edit Calliope Rev 1.3

HaraldRau:
That's the changed for Calliope Mini Rev 1.3! Peter van Epp has done the most work.
The graphics were checked and adjusted. The component was tested by us. I have copied the components in addition to the existing version of Calliope. This version is an update of the existing version Calliope Rev 1.3!

I hope it's the right way to copy the Fritzing-Parts in the Branch nightyParts


KjellMorgenstern:
As I understand Calliope_mini_V_1_3 should replace CalliopeMiniRev1.3.
I don't see a point in merging two versions (with a confusing similar name) into master.

HaraldRau:
Yes, in any case. I just do not know Github so well and I do not know if that works.
It would be great if the new version can replace the old one - should I have had to work differently?

HaraldRau:
Must I rename the Files from Calliope_mini_V_1_3 to CalliopeMiniRev1.3? So I will do this!

HaraldRau:
I delete the other Part from the Calliope Mini.

HaraldRau:
Yes, I had the git-client on nigthlyParts and not on my fork. That was my fault. I will change soon.

KjellMorgenstern:
Thanks a lot, looks great.

HaraldRau:
Thanks, but I still want to create the version of the old Calliope Mini V1. It only has 5 PIN headers and another interface for self-developed shields.

I think next month I will do that.

PR #187, opened 2019-05-02
Enable fzp error checking in CI

KjellMorgenstern:


.travis.yml
```diff
@@ -0,0 +1,22 @@
+language: go
+
+go:
+  - "1.10"
```
ovidiub13: this is an old version of go

.travis.yml
```diff
@@ -0,0 +1,22 @@
...
+
+install:
+  - echo "install fzp tool"
+  - go get github.com/fritzing/fzp/bin/fzp
+  - echo "fzp succefully installed. Version:"
+  - fzp -v
+  - go get github.com/fritzing/fzb/bin/fzb
```
ovidiub13: why are you doing this twice?

core/23LCV1024_1.fzp
```diff
@@ -1,5 +1,5 @@
 <?xml version='1.0' encoding='UTF-8'?>
-<module moduleId="23LCV1024_24357530b44412a899aa27521bf5766d_1" referenceFile="generic_ic_dip_8_300mil.fzp">
+<module moduleId="23LCV1024_24357530b44412a899aa27521bf5766d_1" fritzingVersion="0.12.34" referenceFile="generic_ic_dip_8_300mil.fzp">
```
ovidiub13: Do we have some automated way of updating this number? What exactly does this number represent?

.travis.yml
KjellMorgenstern: Yes, the naming is a bit dangerous. It is fz**p** and fz**b**, so two different validators

ovidiub13: oh, sorry about that.

core/23LCV1024_1.fzp
KjellMorgenstern: Mainly in modelbase.cpp:120 ff. Parts without version are treated as parts < 0.4.2 . The code is a bit tricky, because the greater than has its operands in inverse order, plus inverse of its return value is used. So, having no version has the same effect as having a part that was designed pre 0.4.2. I choose 0.12.34 as replacement as hint that the version is set as a fix.

.travis.yml
KjellMorgenstern: It comes with ubuntu 19.04 by default. Also, the code survived jumping from 1.5 to 1.10 without any problems, so I think it is not very critical.

paulvollmer:
@KjellMorgenstern awesome work!
@el-j can you check the fixed `.fzp` parts?

.travis.yml
paulvollmer: i'm thinking about creating a release at the `fzp` repository to use a specific version for the ci tests.
instead of running `go get ...` we can do a `wget to fetch the binary for the travis-ci machine.

then we cannot break the ci for fritzing parts if there are ongoing changes at the `fzp` cli tool.

ovidiub13: Why aren't we running this in Docker? We could then use what ever version of Go we want.

KjellMorgenstern: docker would be overkill, and quite inconvenient to work with. Are there any issues I overlooked? 1.10 seems to work quite smooth, but of course we can add 1.12 to the CI if you want.

KjellMorgenstern: It would make more sense to test different go versions on the fzp and fzb repos, though.

paulvollmer: i think it is the best to use a prebuild binary for the tests at the fritzing-parts repo. then we can change the travis language to `shell` or something. 

@KjellMorgenstern testing `fzp` and `fzb` against different go versions is a good idea. i already setup tests for go 1.5, 1.6 and the latest. need to update the travis config to test the other versions also

PR #188, opened 2019-05-07
lm239

brahamcrackercrunch:
LM239 Quad Comparator


KjellMorgenstern:
According to http://www.ti.com/lit/ds/symlink/lm2901.pdf , the inputs from 4,5 should go to pin 2. At least in the schematics view it seems that pin 1 and 2 are swapped.

KjellMorgenstern:
Hi! The part looks great now. I squashed the PR to keep the git history clean, and minified the icon, see PR #193. Could you please confirm/comment on #193, I will then merge it.

PR #190, opened 2019-05-10
Calliope Mini V 1.0

HaraldRau:
This is the previous version of the Calliope Mini. In this version there is a PIN header with only 5x1 pins. As a result, extensions do not fit on both versions.


KjellMorgenstern:
Please keep the git history clean and readable. 
This PR should be rebased (you reused your branch without rebasing it, so other commits which  are already in appear again) . Further, there are lots of commit with the same commit message, which is difficult to read and defeats the purpose of this messages. 
If you want, I can squash your PR into a new one, to show what I mean.

HaraldRau:
I was already wondering about the old commits. I thought they were done.

I will close the RP and delete the fork. When I set a new fork on fritzing-prats would the RP work?

KjellMorgenstern:
That sound a bit more complicated than it needs to be, and I am not sure if this will solve the organization of the commits.

HaraldRau:
O.K.,  can you squash your PR into a new. I have no plan, how it works?

KjellMorgenstern:
I have created PR #191 from PR #190.
Are you working for Calliope?

HaraldRau:
No, I am a teacher who is training other teachers and would like to introduce Calliope in school.

I think that's a very good way for children to work with digital technology! I have tested it with children in elementary school. They were excited - i am also!

i am from Brandenburg, around Berlin!

HaraldRau:
Oh, thank you for the RP. 

Oh, but I know Joern Alraun. He knows about my work at Calliope for Fritzing.

KjellMorgenstern:
I loaded the Calliope Mini V 1.0 in my local build of Fritzing, looks all fine. Can you comment into PR #191 ?

PR #191, opened 2019-05-10
Calliope Mini V 1.0

KjellMorgenstern:
Previous version with 5x1 pin header
Vorhergehende Version mit PIN-Header 5x1


HaraldRau:
That is very nice. What do I have to do next?

PR #195, opened 2019-07-05
Fix xml errors in part definitions.

jensh:
Removed some duplicate attributes and added some missing spaces.

scripts/genmoduleidfzphash.py: xml.dom.minidom.parse() had problems with them.

### DIFF
--- a/core/Bean_revE.fzp
+++ b/core/Bean_revE.fzp
@@ -158,9 +158,9 @@ Zero wires. Infinite uses. What will you build next?</description>
 <breadboardView>
 <p layer='breadboard' svgId='connector81pin'/></breadboardView>
 <schematicView>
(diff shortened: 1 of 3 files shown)


jensh:
i revert the changes in obsolete.

KjellMorgenstern:
Sorry, I overlooked, PR should go to nightlyparts, not directly to master. I reverted the change and created a new PR

KjellMorgenstern:
Fix was now merged with #196 to nightlyparts , no conflicts. Also, I updated the settings, so nightlyparts is the github default. Sorry about the mixup.

vanepp:
"Looks good to me, but I wouldn't touch anything in the 'obsolete' folder. @vanepp What do you think about this?" I agree that modifying the obsolete folder might not be wise. I don't entirely understand how the obsolete mechanism works and how sensitive it is to changes in previous xml. Otherwise the changes are good, as they change technically wrong xml (which Fritzing accepts) in to correct xml.

PR #213, opened 2019-08-18
Add voltage regulators from dmantione

KjellMorgenstern:


KjellMorgenstern:
Open issue: All parts have the same, wrong icon. The would show as LM2574 in the parts inspector.

PR #214, opened 2019-08-28
Corrected versions of all the FIRST robotics parts

vanepp:
Corrected versions of all the FIRST robotics parts to replace pull request  141. Because I haven't done bins before I replaced the parts data base on 0.9.3b with this repo and did a manual database update (which failed on 0.9.4 prerelease, I'll poke a that and file a bug if it isn't my error) and verified all the parts and the new bin load correctly. There are some minor changes (part file names) from the review set posted in the Fritzing forums, but the parts are all the same as there.


KjellMorgenstern:
Approved. I rebased on latest develop branch, minified the icons and listed the new parts in the commit message. Pushed to develop branch with commit b0235d3 .

PR #222, opened 2019-09-21
Arduino sensor shield V5, gear-motor, ky_006, shaft-encoder, yl99 parts

vanepp:
Create a new hopefully non conflicting pull request for the listed 5 parts. I ended up deleting my fritzing-part fork and re forking it because I had managed to corrupt the original repositories. I did get a git fetch upline and merge to update my local repository, but pushing it to my fritzing-parts repo on github didn't clear the conflict on my original pull request. Hopefully this will create a proper pull request with no conflicts and a git fetch upline and merge in future will keep my local repo in sync in future (we will see when someone else makes a change to parts I expect).


KjellMorgenstern:
This is the 5.1 version of the shield, but in the fzp file the version tag is 
 <version>4</version>
Is that intended?

@vanepp: I updated this PR by merging in the latest changes (go version had to be increased, because some package was not available anymore for go 1.10, and I minfied the icons files). So you might want to sync first if you have additional changes.

vanepp:
"This is the 5.1 version of the shield, but in the fzp file the version tag is
<version>4</version>
Is that intended?"

Mostly due to inattention, I just left it at what it was set to in the original. It isn't clear what the version field is supposed to be, I suspect since this is the first version of the part it should probably be 1 because I suspect it is supposed to be the version of the Fritzing part. It is used by the obsoleting mechanism, via a special value, according to the parts file format (I haven't yet gotten to figuring out how obsoleting works, it has a number of options in a number of places). I'll try and sync my repos which will be a good test of whether I can in fact keep them properly synced, and change the version numbers to 1 as I suspect most folks do as I did and just use what is there. What the version number should be is yet another thing to research when I get time :-) . If the numbers should be starting at 1, then the check script should be checking for that. The hard part may be figuring out if there is already a version 1 part, but a scan of the parts repository should pick that up in most cases I think. How do you minfied (minify as singular?) the icons? That is something the parts script could do as well. I've been thinking of making the default (overrideable by a command line flag, for when you want a separate icon file as some parts do) to make the icon file the breadboard file as that is one of the options (and the usual one) in parts editor, to save space. We already have the breadboard svg, we don't need a second copy as the icon file and Fritzing is happy to use a breadboard scg as the icon. Reducing the size of the icon would be a performance gain in rendering as well though and thus well worth doing. I'd also like to have an option to specify a pcb svg from core (but that would take a code change to be able to specify that you want a standard footprnt) as we have many redundant copies of DIP(number of pins) pcb svg files when we really only need one.

KjellMorgenstern:
@vanepp 'git status' will tell you the local status. 
To get remote changes, a possible command is "git fetch --all"

PR #230, opened 2019-12-31
Heltec WiFi LoRa 32 module

UniquePete:
Offering Heltec WiFi LoRa 32 module for inclusion in fritzing-parts


UniquePete:
This is my first attempt to submit anything to GitHub, but this does appear to be a Fritzing-specific issue. Either way, can anyone offer any advice on what's gone wrong here (I'm assuming that this comment will be tied to the failure report that I received) and, perhaps more importantly, how I can rectify the problem?

Thanks.

KjellMorgenstern:
The failure is related to our toolchain, it seems so package from the Go language received an incompatible update. 
But you should remove the .DS_Store files, they are from MacOS and don't have anything to do with the Fritzing parts. As a rule of thumb for using git, when calling 'git add' give the exact files names, almost never use wildcards or add complete directories.

UniquePete:
Thanks for that, but what do I do now? Just try again? How will I know when the Go language package is OK again?

As for the .DS_Store files, yes I saw them appear (and figured that they were unnecessary) but I was using GitHub Desktop and I think that just clones the whole directory, no matter what I try to do. If I were to delete the .DS_Store files, I think they’d just get put back again… or is there something I can do to stop that?

This is the first time that I’ve done anything on GitHub, so I’m a complete novice. I thought the Desktop app would make things easier, but I can work out the CLI side of things if need be.

Pete

KjellMorgenstern:
Hello, I fixed the broken test script (the CLI parser API had changed upstream), and all the tests now pass. However you overlooked some of the .DS_store files. 
Also, I saw you published another version on the forum. Or is that the same version as in this pull request?

UniquePete:
No, I think the forum version just has a slightly modified Breadboard/Icon image following the creation of a part for the V2 module. I'll get all of the parts I posted on the forum over here in the next couple of days.

KjellMorgenstern:
Closing, as this PR doesn't seem to contain any files except .DS_store garbage data.

PR #231, opened 2019-12-31
Heltec Wireless Stick Lite module

UniquePete:
Offering Heltec Wireless Stick Lite module for inclusion in fritzing-parts


UniquePete:
Is there an actual problem with this part (after the CLI parser issue was resolved)?

PR #235, opened 2020-02-11
replace LBT2088AH part with generic family of led matrix parts

mMerlin:
Handles 4 different sizes of 8x8 generic LED dot matrix displays, with cathodes on either rows or columns, and with standard circular pcb pads, or oblong pads used for IC chips.  Part number can be set using inspector.

Parts update configured for the obsoleted LBT2088H specific part.  Simple update test by @vanepp worked.  More complex test messed up the pcb view.  Investigation in progress.  Sketch corruption suspected from the work flow used.


mMerlin:
@KjellMorgenstern I assume you get notifications about build failures.  `cannot find package "github.com/cpuguy83/go-md2man/v2/md2man"` looks like something is missing, but travis is not in my skill set.

KjellMorgenstern:
Hi, can you rebased against the develop branch?
Also, maybe i have overlooked documentation about the nightlybuild branch which needs to be updated.

mMerlin:
nightlyparts reference: https://github.com/fritzing/fritzing-parts/blob/develop/CONTRIBUTING.md
I've never done a rebase.  Time to learn something new.

KjellMorgenstern:
If you are lucky, it is as simple as choosing "develop" as base branch above. I'll try.

KjellMorgenstern:
Ok, doesn't rebase automatically. Still, it will be fairly simple, let me know if I can help.

mMerlin:
@KjellMorgenstern I think that's got it.  After some reading and cross checks, it amounted to:
```sh
git checkout --track origin/develop
git merge upstream/develop
git rebase --onto develop upstream/nightlyParts led_matrix
git push --force origin
```

KjellMorgenstern:
Did you execute it? With the push, changes should have shown up here automatically.
My approach would be:

```
# get PR235 to my local system
hub pr checkout 235

# do the rebase (cherry-pick would also work, I often find it more inuitive to type, but it is more commands)
git rebase --onto origin/develop origin/nightlParts

# have a quick check
git show HEAD --name-only

# publish it back to PR235
git push -f
```

Of course those command might vary, depending on how you current local repo looks like.

mMerlin:
Yes I ran that.  I do not recognize that hub pr command.  That **look** like something specific to github pull requests.  Guessing is supposed to do a checkout to match the existing pr, to base the following commands on.  The git show command reports exactly the expected new and obsoleted part files.  The pr here shows fritzing:develop.  I guess from your shortcut attempt.  I don't know what more I can do or check.  I have no knowledge about what the conflicting files being reported.  Other than a guess they are related to switching from nightlyParts to develop.

Do I need to create a new pull request from the rebased branch?

KjellMorgenstern:
"git push --force origin"
I think you where on the develop branch. In that case, you have updated your fork of develop, but not the PR. This might cause conflicts when you later sync upstream develop. To update the PR, changes need to go to the led_matrix branch.

KjellMorgenstern:
Yes, hub is a CLI tool for using github from the command line. It can create issues, forks, and manage pull requests.

KjellMorgenstern:
Please don't start over. For the learning effect, you could try to figure out how to solve it without starting over, or opening a new PR. But to save time, I can also just push my version :-)

mMerlin:
For 'learning', to do the rebase, I first cloned my local repository, and did the work there.  In the original local repo, `git status` now says
```sh
Your branch and 'origin/led_matrix' have diverged,
and have 41 and 64 different commits each, respectively.
```
So it sure looks like the rebase was pushed.

Also, for learning, I don't know what I need to know, to figure out what is/went wrong.  Where to start.  Can you provide a pointer?  I deal with manuals and documentation quite well, but need some idea of what the topic is to get started.

If starting over, it would be easy to create a new branch off of develop, and reapply the changes there.  Already did that once, when I saw the (incorrect) reference to nightlyParts.  It is a single commit.  Just need to do the 5 `git mv` commands, then rsync the new and modified files.  All in my notes, because I was carefully tracking the steps done.  :slightly_smiling_face: for learning

KjellMorgenstern:
You didn't push anything to "led_matrix" yet.

mMerlin:
```
git rebase --onto develop upstream/nightlyParts led_matrix
git push --force origin
```
From my research, I thought that the rebase would leave me in led_matrix branch, at the end of the rebase.

KjellMorgenstern:
One command further back "git checkout --track origin/develop" ...

mMerlin:
That was to make sure that my local develop branch was synchronized with upstream before doing the rebase.  Since that was an input to the rebase.

Collecting information to show the current state, I found the problem.  When I cloned the local repo, I did not change origin back to github.  So

```
% git remote -v
origin	/home/phil/development/FritzingProjects/FritzingParts/mmerlin/fritzing-parts (fetch)
origin	/home/phil/development/FritzingProjects/FritzingParts/mmerlin/fritzing-parts (push)
upstream	https://github.com/fritzing/fritzing-parts (fetch)
upstream	https://github.com/fritzing/fritzing-parts (push)
```
And the push I did was back to the other local repo.  Sigh.  Let me think a bit on how to continue from here.

KjellMorgenstern:
Regarding how to learn git... i think there are a few basic commands to learn to get started, like the happy flow: clone, add, commit, push. 
Then git pull and git mergetool to resolve conflicts.
Then just go through what you are doing now several times, not giving up (deleting the folder and cloning again counts as giving up ;-) ).
Using it on a daily basis for more than a decade certainly helps, too

mMerlin:
Been doing things like that.  Adding in creating new project branches.  Messed up because I got off of the main path, into things I had not used before.  I cloned here, intending to keep a snapshot point, incase I messed up, that was what actually caused the problem.

```sh
git remote add ghub https://github.com/mMerlin/fritzing-parts
git fetch ghub
git co led_matrix
git push --force --set-upstream ghub HEAD
```
Seems to have done it.  I see a new message here about the force push.

mMerlin:
> Hi, can you rebased against the develop branch?
> Also, maybe i have overlooked documentation about the nighlyparts branch which needs to be updated.

Already handled, but before I updated my local repository, 
```
git co develop
grep -ir "nightlyparts" *
```
found it, and currently
```
git co master
grep -ir "nightlyparts" *
```
shows the same reference still in CONTRIBUTING.md, plus a couple of lines in README.md

Nothing found in the fritzing-app wiki, and the fritzing-parts wiki is currently empty.

vanepp:
While I'm glad to see github doesn't only bite me when I try and update parts (and I haven't yet tried the git commands posted in the forum to see if I can sync my github parts repo to the real parts repository), I think we need a simple howto (which I can write if I ever figure out how to do this) on keeping a cloned repository on github in sync with the master. This is such a common task that I'm surprised there isn't a tutorial on github (or else where) but I haven't been able to find one. The usual ones are from a local to a remote repository, none with a local repository, a writable by me repository on github and a cloned repository that you are making pull requests against. While at present not many other than me are making parts, even if they were I suspect like me they may end up posting in the forum because they can't figure out how to make the pull request on github.

KjellMorgenstern:
@vanepp Have you tried the keyword „fork“  „How to keep a fork in sync with an upstream repository“?
Also, one advise would be to not have long lived feature branches. Create a branch, like „led_matrix“, make a pull request and delete it. If you keep a branch, you will have to rebase it frequently.
Maybe one thing that is noteworthy is that I ask pull request to be fast forward commits on the develop branch. Usually you don’t notice. The difference is for long lived branches, that you should frequently rebase instead of merge.

KjellMorgenstern:
@mMerlin At a first glance, this PR looks great. I also tried a simple „obsolete“, and it worked. I would like to investigate the issue (disconnecting) described in the forum a bit more.

KjellMorgenstern:
After updating the obsolete part:
image

KjellMorgenstern:
We need to figure out if this is an issue somewhat triggered by the part, and if there is a workaround for that, for example by avoiding an offset of the connectors.
Assuming the part is sane, the issue should also be fixed on the application side.

mMerlin:
There is some updated information on the forum at https://forum.fritzing.org/t/part-update-pcb-view-mess-possible-corruption-in-sketch/8455/5 if you have not seen it yet.  With further testing and results done with the update process for that part.  I recognize issues in your screen shots as matching what I saw and documented there. You did not see (or at least show here) all that I found.  With the symptoms, the (my) current estimate is (at least) bugs in the code.  Until we have some simpler cases that work, and can be made to fail in specific ways by simple changes, that is an assumption.  As a start though, we could regress the parts database (checkout old commit) to a point before some other part was obsoleted, create a new test file with that part, and try updating.  Time to dig out some git commands to locate commits with files being moved to obsolete.  Preferably something with more than a couple of pins, and with replacedBy attributes for (some of) the connections.  If that has the same kind of symptoms, it should add weight to the theory that there are bugs in that block of code causing the problems.  grep found a few candidates.  The diodes might be useful simple case to use for further testing, if they fail.

```sh
% grep -m1 "connector.*replacedby" *
0223_0eb44bbb4c4f50c21567b8c79110a3c3_5.fzp:  <connector type="male" id="connector130" name="Enable" replacedby="ENABLE">
Arduino Ethernet Shield.fzp:        <connector id='connector15' name='D0RX' type='female' replacedby='RX/D0'>
Arduino Fio.fzp:        <connector id="connector15" name="GND" replacedby="connector23" type="male">
arduino.fzp:        <connector id="connector15" name="D0RX" replacedby="D0/RX" type="female">
Arduino_Leonardo_Rev3.fzp:<connector id='connector84' type='female' name='IOREF' replacedby="connector85pin">
Arduino LilyPad.fzp:        <connector id="connector15" name="D0" type="female" replacedby="connector49">
arduino_mega.fzp:        <connector id="connector1" name="GND" replacedby="connector237" type="female">
Arduino Mini.fzp:        <connector id="connector15" name="+5V" replacedby="connector56" type="female">
Arduino Mini Pro.fzp:        <connector id='connector15' name='BLK' type='female' replacedby='GND'>
Arduino Nano3.fzp:<connector id='connector55' type='male' name='D12/MISO' replacedby="connector30">
Arduino Nano_v30.fzp:        <connector id="connector15" name="D12" replacedby="D12/MISO" type="male" replacedby="connector30">
Arduino-Pro-Mini-v11.fzp:<connector id='connector14' type='male' name='DTR' replacedby='DTR'>
Arduino_UNO_R3.fzp:        <connector id="connector28" name="reserved" type="female" replacedby="N/C">
avr_isp_10_pin.fzp:        <connector id="connector3" type="male" name="sck" replacedby="connector6">
RJ45_TRXCOM.fzp:        <connector id="connector3" type="male" name="Transmit CT"  replacedby="connector3">
SMD_mega2560.fzp:<connector id='connector0' type='male' name='(OC0B) PG5' replacedby='connector0'>
sparkfun-discretesemi-diode-1n4148.fzp:<p layer="breadboard" svgId="connector0pin" replacedby="connector1pin"/>
sparkfun-discretesemi-diode-hv.fzp:<p layer="breadboard" svgId="connector0pin" replacedby="connector1pin"/>
sparkfun-discretesemi-diode-pth.fzp:<p layer="breadboard" svgId="connector0pin" replacedby="connector1pin"/>
```

For the "simple test case" path, one way is to create a perfectly good part in a test branch (a clone of an existing good part), create a fzz with that, obsolete the part with what it was clone from, and test and update.  Assuming that works, deliberately make changes to the obsolete part, to make specific things different from the replacement, and repeat.  Saving (part file) snapshots of any failing cases (regression test cases).  Since some of the current issues seem to stem from (at least breadboard) (mis)alignment, some of those tests will need to change the view graphic images.  Not necessarily the pins/pads/connections.  Changing the outer graphics will change the bounding box, and (I think) placement coordinates.  I think it would be better to use the corner of the bounding box for the pin/terminal centers as the placement origin, but symptoms indicate the corner of the image graphics is used.  Which has varying distances from the pins, and is not even close to 0.1in multiples from there.

> We need to figure out if this is an issue somewhat triggered by the part, and if there is a workaround for that, for example by avoiding an offset of the connectors.

Do you mean the use of transform="translate()" in the new part images? or something else?

If that is a factor, it might be for the obsoleted part as well.  But with these part images, it is a fairly easy bulk edit.  As long as do not need to get rid of all of the wrapper transforms that adjust the local origin to make most coordinates into nice easy multiples of 0.1 inch.  That would take more work.  And be a significant issue with images created with inkscape.  These hand coded svg images are cleanly structured, making it at least practical.  There are not a lot of transforms, but still need to calculate new coordinates for everything wrapped inside.  I could probably write a program to do that, but that would take time too.  More time, but it could become another "conversion" tool, if actually needed.  Better to fix the code doing the update.  Assuming that is the real cause.

I **COULD** redo this pr, to not obsolete the old part.  I could get more practice with rebase :slightly_smiling_face: There is no direct conflict.  Just prefer to not have the Inspector keep offering the key words for the old part in the drop downs.  But if we can not move forward with the symptoms being seen until they are resolved (tracked to a root cause that is not either the new part, or the way I obsoleted the old part), then that is an option.  It can be obsoleted as a separate pull request when enough information is available to permit it.

KjellMorgenstern:
No need to touch the transforms, 
```
 width="1.51in"
 height="1.51in"
 viewBox="0 0 151 151"
```
I think this is perfectly good.
The issue is probably to be found in the application, but if there is a workaround, it should be used, so that it works well with Fritzing version < 0.9.5  (so, all versions out there...)

You have a 0.2.2 fritzingVersion.  That should be > 0.5 it think, since that is the latest version so are where fritzing tries to apply some magic to work around part format issues. From a guts feeling, it should be 0.9.4, as this is the version were are testing it with the most.

The referenceFile tag seems to do no harm, but it is unused and kind of confusing.
With those minor changes fixed, I think we can integrate the PR.

The obsoletion issue should be dealt with in an new ticket,  would be great to reproduce it with somehting simpler like a diode.

mMerlin:
@vanepp 
> While I'm glad to see github doesn't only bite me when I try and update parts (and I haven't yet tried the git commands posted in the forum to see if I can sync my github parts repo to the real parts repository), I think we need a simple howto (which I can write if I ever figure out how to do this) on keeping a cloned repository on github in sync with the master. This is such a common task that I'm surprised there isn't a tutorial on github (or else where) but I haven't been able to find one.

Here you go :smile: https://github.com/fritzing/fritzing-parts/wiki/contributing-new-part-process

@KjellMorgenstern Could you scan that as well.  It is intended to be an end to end happy path.  I know some git, but I don't have 10 years of daily exposure.

Ah! for offsets, you meant the mapping between physical units and internal coordinate.  I was quite careful with that, getting nice clean numbers.

The fritzing version was deliberately set low.  I researched the code to see how it is used.  The way I interpret that, the version number should be the lowest version of the software application that knows how to process the part information.  The current code does alternate processing for less than / greater than specific version numbers.  This does not have any features not in the original part, and afaik, nothing to trigger the existing version number checks.  So I set it the same as the original part, figuring that any version that could handle that part could also handle these.  0.9.4 was where I started as well, until I read the references in the code.

The referenceFile attributes I will be happy to remove.  And ignore the complaint from FritzingCheckPart.py

Are you going to want another rebase, or is another commit enough?  With that description, should I really change the fritzing version as well?

Reproducing the update problems with simpler parts in on the list.  Along with things like the git cycle documentation dumped into the fritzing-parts wiki.  And some actual using of Fritzing for the local robotics society.

vanepp:
I think the " for example by avoiding an offset of the connectors." comment refers to the pins not being exactly on .1in boundaries (and at .1 pitch)  in breadboard and schematic (obviously pcb has to be the exact position.) The " I think it would be better to use the corner of the bounding box for the pin/terminal centers as the placement origin, but symptoms indicate the corner of the image graphics is used. " doesn't appear to be that simple. Fritzing appears to be using a voting algorithm of some kind to set pin origins. The theory in the forums was that pin0 is used as the base, but testing indicates that isn't true. It seems to vote on a center line some how, this shows up when there are 2mm connectors in breadboard, the .1 pins will sometimes not be on the grid. I would expect to be able to measure the offset in x and y, and adjust the svg to move the .1 pins to be on the grid. That doesn't work. Moving the pins in x and y so they should align to the grid (I Was assuming the start point was one of the 2mm connectors and moving just the .1 connectors should be linear) moves, but not on to the grid it shifts to a new position. I tried for a few days with 4 .1aligned connectors and moved them one at a time to a non aligned .1 position expecting as one connector moved all the rest would too. Doesn't happen, more than one connector needs to move before alignment will change. I never was able to figure out what Fritzing uses as the base for the connector coords, another thing on my list of things to look at in the source that I haven''t gotten to yet.

KjellMorgenstern:
@vanepp That doesn't sound like a heuristic (voting algorithm) at all, more like a numeric problem. If you look into it, be suspicious about each "==" comparision. On a computer 2*PI + 3*PI == 5*PI will most certainly return false because floating point errors.

vanepp:
@KjellMorgenstern "That doesn't sound like a heuristic (voting algorithm) at all, more like a numeric problem." Good point! I couldn't find any pattern that worked, and didn't think of floating point round off type problems which are much more likely than a voting algorithm. I'll add that to the bug list for when I get back to that.  Python has a floating point "matches close enough" library, I expect there is something similar in Fritzing for the same reason and I bet it has been forgotten in the code that sets initial connector position. Thanks!

mMerlin:
> but symptoms indicate the corner of the image graphics is used.
@vanepp 
> doesn't appear to be that simple.

I was talking about the position of the part itself, as reported by Inspector.

I was basing my analysis on these observations.
* setting the location of a part in Inspector to a simple number (0,0) does not usually put the connector pins on the .1 grid
* positioning 2 different parts at the same coordinate in Inspector does not align the part connector pins
* for the 2 parts connector offsets are not a multiple of .1 inch.

Take for example a diode and capacitor.  On breadboard view, without a breadboard to get in the way, use inspector to position them both to 0,0.  With align to grid, show grid on, .1in spacing, place a resistor so the right end is just above the left end of the diode connector.  What I see is that the tip of the connector for the diode graphic is horizontally aligned with the vertical .1in grid, but the top of the diode graphics is on the grid line, not the connector.  The connectors for the capacitor are not on either grid, but the left edge of the capacitor graphic is on the grid, where the wire arcs out, and the top of capacitor graphic is on the grid line.  With the resistor snapped in place, look at the coordinate shown by inspector.  -0.415, -0.050.  left 0.415in and up .05in.  The resistor connections are 0.4 inch apart, but that extra .015 is because the graphics extends a little past that.  From the rounded end on the end of the connector line.  The graphic is visually .1in tall, and vertical offset is half that.

The origin (placement point) for each of those cases is the top left of the graphical bounding box for the part.  Which is actually larger than the bounding box shown with a part is selected.  When using align to grid, the connector is snapped to the grid, but the part location is off the grid.  I have not checked, but I assume that the selected part bounding box is based on the SVG viewBox, but the render graphics can extend beyond that (especially for bendable legs), and that graphical rendered bounding box is used to define the origin/anchor point.

If the placement (origin, anchor) point was the top left connection point in the part, the parts could easily be placed in exact positions using Inspector.  WIth rotations that are not multiples of 90 degrees making things harder.  Which would probably be handled by Inspector applying a fixed (calculated per part) offset to the location values.  The delta from the existing bounding box to the centre point of the top left connection in the part.  With rounding so do not get those 0.001 values.

vanepp:
I think I have found a workaround to this problem. I finally thought to have a look at the original part's breadboard svg, and indeed the top row of connectors are misaligned by 0.011in in y. The bottom row are x 0..822 y 02.50 (in) the top row are x 0.822 y 2.061 (in). So I edited fritzing-parts/svg/obsolete/breadboard/red_led_matrix8x8.svg and adjusted group "0.3" (the top row connectors to be at 2.050 in y rather than the current 20.61 (which causes a transform that I would normally have removed, but left in this time to make a minimal change) and saved the svg. Now the LBT2088AH-test.fzz when loaded and updated updates correctly in all three views (at least as far as I can see.) That suggests by doing something bad (modifying the original svg slightly in the obsolete folder) we can fix this problem. My original test part (test.fzz) looks to load fine with the modified svg in obsolete (which I admit is a not very complete test) so old sketches should still be fine. It also looks to tell us where the bug lies, as it appears to be an alignment change in the connectors that trips it. I'll have to think about your post above a bit to comment on it (just saw it as I am posting this.) I think this is normal, the connectors align independent of the view box, except sometimes (in Inkscape at least) if you do not do a resize page to document to update the viewbox you can cause connectors to be misaligned (such as offset 0.05in from the grid in x, y or both even with all connectors aligned on 0.1in centers.) I think this is part of the connector alignment issue that I mentioned above.

mMerlin:
@vanepp So the svg images are not stored in the fz/fzz sketch file?  I sort of thought snapshots of parts used in the sketch were kept with the drawing.  Maybe that is only for custom user parts (fzpz)?  I seem to remember an explicit test when I was doing my initial custom fzpz parts files.  To make sure that a user could display the sketch properly without first loading my custom part that was used in it.  Since that testing worked, I assumed the parts were embedded in the sketch file.  Again, maybe, especially with what you say, that only applies to the user parts.

> I think this is normal, the connectors align independent of the view box

Yes part (grid) alignment in the view will (better be) based on the connector position.  My comment is that the part position shown by Inspector is based on the overall graphic, not the (reference) connector position.  It makes the reported coordinates not very useful for getting the connectors aligned.  In breadboard and schematic views.  For pcb, the actual physical part boundaries are also important.  Which is not necessarily the same as the silkscreen, or other visible graphic boundary.  There is an "outline" id that can be used in (at least) pcb svg files.  I made note of it when I did the initial research for part creation, but have never looked at part that used it.  There are a few that exist, but they are almost all empty.  A quick grep and scan only spotted 4 that were not empty.  Most pcb svg files do not even have it.  But that is what probably *should* be used on pcb view for physical interference checking.  Which is still separate from the anchor position to use for the part.

Did you see the contribution process wiki document?  It includes the git commands to resync the local, forked, and upstream repositories.  The pull request section is a bit vague, because I was not in a position to walk through and document the prompts.  Summary at the end.

@KjellMorgenstern Bump.  Looking for feedback on my comment about what needs changing before accepting the pull request, and how to send the changes once done.

vanepp:
@mMerlin "So the svg images are not stored in the fz/fzz sketch file? I sort of thought snapshots of parts used in the sketch were kept with the drawing. " Not for parts in core. Only parts in the mine parts bin to become (which is local and may not be on other systems) are included in the fzz file. The files from the mine parts bin end up in temp (and having the same moduleId but a different version in the mine parts bin causes problems, another bug :-) ., as Fritzing prefers the mine parts bin and takes a potentially incorrect part (if the versions are different.) 

" My comment is that the part position shown by Inspector is based on the overall graphic, not the (reference) connector position."  There is something odd (perhaps a bug) with the coordinates in Inspector. While working on an accordion interface for someone a while back he needed parts in a particular place on the pcb to match the physical position of magnets in the accordion, I discovered that when units are set to mm you can only set some values (I suspect there is a conversion to other units in there, with a not fine enough granularity), so trying to increase a coordinate by 1.5 mm would in fact increase it by 1.345 not 1.5, and I couldn't get the exact fractional mm I wanted. Another on the long list of things to look at when I have time. 

"There is an "outline" id that can be used in (at least) pcb svg files. " I believe the outline layer is what is used when loading logo files and cutouts in pcb. I think the single path additive/subtractive path generates an outline layer to do the cutout. It also AFAIK sets the border of the pcb. The logo code  has a number of bugs (Kjell fixed 5 or more of them a while back) and is quite difficult to use @opera_night in the Fritizing forums is the expert there and has had the most luck with it. I usually end up with a mess, I don't think I have ever generated a successful cutout.

"Did you see the contribution process wiki document? " Yes I did, I will probably try to figure out how to edit the wiki and add pointers to the only 3 sets of tutorials that are specific to 0.9.3b/0.94, all the rest I know of are for older versions pre parts editor. I also need to try the git commands and see if I can sync my parts repositories which I haven't done yet. 

As to the pull request,  I'll see if I can upload the svg file I changed which makes the update work, as I think it would be useful to include it so as not to break things during the update. 

[red_led_matrix8x8.svg.zip](https://github.com/fritzing/fritzing-parts/files/4196208/red_led_matrix8x8.svg.zip)

It doesn't accept svg files so I added a .zip to the end (without zipping it) and it now appears happy.
This wants to be in fritzing-parts/svg/obsolete/breadboard to make the update happy (and I think not break anything else.) This just aligns the connectors on the .1 grid in breadboard.

mMerlin:
@vanepp 
<  (which causes a transform that I would normally have removed, but left in this time to make a minimal change)
**minimal** change he says.  Maybe graphically, but I tried to do a diff between the versions.  Looks like you ran that file through FritzingCheckPart either before or after editing (with inkscape?).  The original was not 'checked', so way different layout.  So I tried running both the original and the file you provided through FritzingCheckPart, figuring to compare the outputs.  Not very good there either.  There was some synchronization between blocks, but the attribute orders were different.  I **thought** that part of what FritzingCheckPart was doing, was sorting some things to make compares like this possible.  I can visually match the differences shown by diff, (slowly) one at a time, but the raw diff is still too different.  Didn't you find a practical (and safe) way to sort the attribute keys?

I'm going to see if I can replicate the changes you did, without totally rewriting the svg file.  So that git can see what the actually changes are.

This also needs to be tested with an old part / old drawing, to see if it breaks if the upgrade is declined.  I am thinking that might break in the same way the the update broke without adjustment to the obsolete svg.

vanepp:
@mMerlin "minimal change he says. " No, I didn't run it through check part, just Inkscape (although that can be bad enough.) Because Inkscape is written in python, the attribute order usually comes out different as the dictionaries output them in random order. It may also change items it doesn't like that Illustrator did for CSS compatibility (perhaps not always correctly) which may explain the 4 zero length line segments change (because I didn't do anything but move the one row without ungrouping everything which causes a tranform.)  I had forgotten that a diff of an Inkscape modified svg file is usually hopeless. There isn't much Check part can do because lxml does the same thing (and worse yet, I have a custom pretty printer in it which changes the format still more.) I'll poke at whether there is a way in Check part or more likely the pretty printer, which is after lxml is done producing the document, to output a sorted in some manner svg so diff would work against Inkscape modified svgs as long as they were both run through Check part first. That won't fix things like the 4 zero length line segments though  so it may not help enough. 

"brd2svg " is part of Eagle2fritzing which takes an Eagle board file and converts it (with varying success) to a Fritzing part, so it could be the cause (or its original input anyway), although I'm surprised there was an Eagle file for a component, perhaps the author put it on a board to make the part more easily, assuming there was an Eagle footprint for it, although the pin numbering isn't screwed up enough for a usual Eagle2Fritzing board (although perhaps one with a single component would operate correctly, I usually do boards with components that get connector definitions that aren't wanted in Fritzing). The missing pin in the sequence is usually a Parts Editor bug, so there may be multiple causes.  It may also do footprints and I've just never done one though, although brd2svg takes a .brd file as input normally and I think the footprints are a different format. Changing the original file with a text editor to only change "158.682" to "158.288" and only in that one group, should produce the desired minimal change without changing the format, so diff should only find that change. That really would be the minimal change but I didn't think of it: svg edit with Inkscape :-) . If you only have a hammer everything looks like a nail.

mMerlin:
@KjellMorgenstern Do you need anything for from here?  Just busy?  Requested adjustments committed.  I have not seen a response to my reply about the wiki content either.

KjellMorgenstern:
Just busy. It nice that this works now, but I am a bit sceptical about modifying the coordinates in the obsoleted part. How is the offset caused by this compensated, will the copper wire have a minor mismatch? How does this look on a high zoom level?

vanepp:
@KjellMorgenstern  "but I am a bit sceptical about modifying the coordinates in the obsoleted part. How is the offset caused by this compensated, will the copper wire have a minor mismatch? How does this look on a high zoom level?"  

I'm a bit skeptical too, but a couple of mitigating points: it isn't a commonly used part (such as a resistor for instance) so the damage (if any) should be in a small number of cases and in old sketches (as the change is to an obsoleted part.) I think (without having a way to prove it) that the change shouldn't cause any issue on the connection. The reason is the change is very small (in the range of 0.01 in) which should get swamped during rendering (most parts have connections that vary more than .01 in in connector pitch just because of svg scaling and roundoff errors and placement errors by the user.) I believe the wire connection is to the center of the pad, thus 0.01in should be negligible (but note the weasel words "should be" :-)  even at high zooms (which in itself is uncommon.) However all good things to think about before making changes, but overall I think this change is safe enough to make.

mMerlin:
You got ahead of me :)  After the scepticism comment from @KjellMorgenstern , I setup a process to do some more detailed comparison before adding the new parts and tweaked obsolete image, and after when declining or accepting the updated part, for the breadboard and pcb views.  I was just cleaning up the last of that when I got the notification that the merge had been done.  Far enough along to finish up (which of course took longer than expected), and post the result for future reference.  The steps used may be useful for the cases of obsolete part update problems.

As expected, there is a bit of part movement when accepting.  The new part does have different graphics after all.  I also did an extra test to get the part realigned to the grid after accepting the update.  That worked well.  I have a (real) zip file here containing 64 images generated during that.  I don't think you want all of those directly embedded in the comments :)).  It is hard to do side by side comparisons to check for differences here anyway.

[obsolete_fudge_test.zip](https://github.com/fritzing/fritzing-parts/files/4258743/obsolete_fudge_test.zip)

I started from (a copy of) the "LBT2088AH test.fzz" file that has been used previously here.  All image files in that zip start with "test".  That is followed by "origin", "marker", "decline", "accept", or "realign" for images from the original file, a version with marker wires and traces added, the marker version after declining the update, the marker version with update accepted, and finally after realigning the parts to the grid.  The next part of the name is "00" for svg export and fit to window screen shots.  Other content here identifies the section of the view zoomed into (maximum 3000%), and ending with "bb" or "pcb" for the view.

Summary

As (I at least) expected, there were no visible changes on the pcb view when declining the update.  The fudge used only changes the image for the breadboard view.

Reference pcb markers at pin 16
test_markers_pin16-pcb

When declined at pcb pin 16
test_decline_pin16-pcb

Accepting the update showed visible movement of the pins on the pcb view.  Enough to cause drc to fail in many cases.

Reference pcb markers at pin 1
test_markers_pin1-pcb

When accepted at pcb pin 1
test_accept_pin1-pcb

When accepted at pcb pin 16 (same reference as for decline)
test_accept_pin16-pcb

Realigning (snapping) back to the grid got rid of most of the position change

Realigned pcb pin 1  (same reference as for accept)
test_realign_pin1-pcb

Realigned pcb pin 16 (same reference as for decline)
test_realign_pin16-pcb

For Breadboard view, when declining the update, the first row of pins, and the part graphics did not move.  The second row of pins shifted slightly, as expected, since the patch explicitly changed the offset between the rows of pins.

Reference breadboard markers at pin 16
test_markers_pin16-bb

When declined at breadboard pin 16
test_decline_pin16-bb

When accepting the update, the breadboard image shifts a little relative to the pins, as well as the pins shifting.  Part of the image rebuild.

Reference breadboard markers at pin 1
test_markers_pin1-bb

When accepting at breadboard pin 1
test_accept_pin1-bb

When accepting at breadboard pin 16 (same references as decline)
test_accept_pin16-bb

After snapping the part back to the grid, both the graphics and the pins are a lot closer to the original position.

After realign to grid at breadboard pin 1 (same reference as accept)
test_realign_pin1-bb

After realign to grid at breadboard pin 16 (same reference as decline)
test_realign_pin16-bb

The exported svg images should be able to be used to zoom in even further, to check for [mis]alignment.  Although, with the tools I use (eog), the svg files first need to be manually edited, to change the global width and height values.  eog only goes to 2000x, and can crash near the limits.  Multiplying those main 2 dimensions by 100 or 1000 avoids the problem. (change for example "4.3218in" to "432.18in").

Here are the sketch files used, renamed to append .zip.  Note that these are backups, so the .fz file inside all 3 .fzz is the same.  "LBT2088AH test.fz".  The first (-a) is the original test file I started with.  The second (-b) is a backup of that with marker lines and traces added, before switching to a parts database with the new and obsoleted parts.  The last (-c) is after updating and realigning the parts to the grid.

[LBT2088AH test-a.fzz.zip](https://github.com/fritzing/fritzing-parts/files/4258824/LBT2088AH.test-a.fzz.zip)
[LBT2088AH test-b.fzz.zip](https://github.com/fritzing/fritzing-parts/files/4258825/LBT2088AH.test-b.fzz.zip)
[LBT2088AH test-c.fzz.zip](https://github.com/fritzing/fritzing-parts/files/4258826/LBT2088AH.test-c.fzz.zip)

KjellMorgenstern:
Ok, so it could be an issue in certain cases. But then, replacing the circular pad with a rectangular pad and other effects might also cause those shifts, and the desing rule checker will alert. Thanks a lot for your patience, I hope this process will get much faster and more fluent.

mMerlin:
Everybody is learning what is needed.  Rough edges are to be expected.  That rectangular pad is only seen when accepting the update, which already has the warning to check all views. The main concern (my perspective) was to not break existing sketches that did not do the update.

Probably should expand the documentation about the repository update (merge) process, to provide guides for testing under certain conditions.  Like walking a (to be) obsoleted part through the decline and accept cycles.  And for cases like this, where additional adjustments are needed to the obsoleted part to get the Fritzing part update process to work properly, the extra tests to check the likelihood that declining the update will break existing sketches.

A note for this case: the breadboard issues were not visible if wires were connected directly to the part terminals.  The wires moved with the change.  The problem showed up there when the part was connected to a breadboard.  In that case, the discrepancy broke the connections, cascading problems over to the pcb view.

vanepp:
@mMerlin "A note for this case: the breadboard issues were not visible if wires were connected directly to the part terminals."

Assuming the tiny breadboard, that may be because the breadboard terminals are slightly misaligned on the .1 boundaries (true of several of the variant breadboards and on my list for the parts clean up when I get to it.) Just having some documentation on obsoleting is a big step forward though!

mMerlin:
With the row spacing, the part would not fit on a single breadboard.  Each row was on a different breadboard.  With the spacing issue in the original part, one of the breadboards had to be moved off of the .1in grid alignment to make the connections in the first place.  So it could also have been the delta between the breadboards that triggered the problem.  It was close enough to get the original part to connect, but not close enough for the correct part. Edge cases.

PR #237, opened 2020-02-15
Nightly parts

A-J-Bauer:
Added 4 'standard' breadboard / prototype board parts.

Sizes 2x8cm, 3x7cm, 4x6cm and 5x7cm.
These are available from various vendors.

I use these for parts placement before actually soldering then as a template/reference and later on for documentation purposes.

If you think this is useful and this makes it into Fritzing, I could provide more boards if desired (the process of creating the files is more or less automated via a console app).


KjellMorgenstern:
Thanks for those parts. I am not in a computer to be able to actually check this:  Did you take into account that the size of perfboards in Fritzing can be configured? What properties do those parts have on top of already available perfboards?

A-J-Bauer:
Hello, I’m aware that the size of the perfboard can be configured. I was using this before.
It’s ok but there is room for improvement (color changeable would be one improvement)

The thing is I usually have those ‘standard’ prototype boards lying around and don’t want to count the rows and columns when trying one of them out for a project.
Now I can just select the “breadboard” size that’s printed on them like in this picture:
[cid:image001.png@01D5E3DE.69DCA390]

It’s a lot faster.

Also with the perfboard it is hard to see resistors, especially when standing up (I wish I could change the body color of those by the way, mine are usually blue):

[cid:image002.png@01D5E3DF.04075820]

Another thing is that I want to see the indication of the mounting holes in order not to put things too close to them.


Finally I like to have the board resemble my actual board as much as possible, makes soldering, documentation easier for me (I usually don’t go all the way to PCB even though I have a mill)

[cid:image004.png@01D5E3E0.A77AA5B0]

I won’t feel sad if this doesn’t make it into core 😊 just thought I try to share.

Regards
A.J.

KjellMorgenstern:
This was autoclosed when I deleted the nightyparts branch. New PRs should go to to the develop branch, but it seems it takes some time until word get around.

Maybe a variant with mounting holes 
and a green variant 
could be added to the magic parts
see file 
./resources/parts/svg/core/breadboard/30.20perfboard.svg
./resources/parts/core/30.20PerfboardModuleID.fzp
in the fritzing-app project.

mMerlin:
> Also with the perfboard it is hard to see resistors, especially when standing up (I wish I could change the body color of those by the way, mine are usually blue).

Did you mean body color of the resistors?  If so, the body color for 4 and 5 band resistors is different.  Select a resistor, in inspector change from 4 to 5 bands.

Screenshot from 2020-02-15 16-10-06

PR #240, opened 2020-03-05
Heltec WiFi LoRa 32 V2

UniquePete:
Heltec WiFi LoRa 32 V2 development board


KjellMorgenstern:
Hi Pete,
I loaded the part and checked it, very nice work.
I think it is a good idea to increase the overall font size in Fritzing (the design had 75dpi in mind, not the >100dpi knowadays). Only I was worried about the style tags in the description, as that looks a bit odd, and I am not sure how well those are supported by Qt. I also added links to the manufacturer (helps users both in sourcing the part and seeing if it is available). 
What do you think?

UniquePete:
Thanks Kjell. This is all very new to me, so I'll accept any guidance that anyone has to offer. All I did was create the .svg files in Illustrator, using the templates provided as a base, and let the Parts Editor do everything else. I did take some advice early on in relation to the file structure and how to 'work that  through' the cleaning and checking process and pretty much all I did there was fix up file names.

I thought I had a link to the Heltec website in there, but I guess that might have been lost along the way somehow.

Please make any changes to any of the files I've just submitted that are necessary. I'll watch what people do, if that's the way things work, and learn for the next time. If I have to work everything out for myself, it'll take a little longer.

UniquePete:
Well, technically, I believe that the CR2032 battery is what is generally described as a Lithium-metal battery. Li-Po batteries are a type of Li-Ion battery. All, at one level or another, depend on the chemistry of lithium, but in different ways. A Li-Po battery uses a polymer electrolyte, but this fact still does not dictate the chemistry of the battery, nor identify characteristics such as its compatibility with the [re]charging circuitry involved in the present case.

If one wanted to be pedantic, I believe that "Li-Ion" would be the more [generally] correct description, although I don't belive that a "lithium-metal battery management system" would make any sense, as Li-metal batteries are, as you note, disposable items (i.e. not rechargeable). I would suggest that the description of a "lithium battery management system" necessarily implies that the battery in question is rechargeable, and thus of Li-Ion technology, but I am happy to amend the reference and check my description of the other Heltec modules, as they all incorporate Li-Ion battery management circuitry.

Just for the record, the current wording is lifted directly from the manufacturer's product description.

KjellMorgenstern:
Ok, no worries. 
I want to take the time to review this and the other pull requests again sometime soon. Do you have more parts in the queue, or are some of the PRs not yet ready?

UniquePete:
With the exception fo this particular matter, I think all of my pull requests are ready now—at least, I think they cover off the various discussions we've had and are all consistent with each other.

I do have some other parts, but they are really only for 'breadboard' usage (i.e. they wouldn't be PCB components and at a schematic level the inputs and outputs are trivial), so I wasn't goiong to bother submitting them formally.

KjellMorgenstern:
In the schematic view, for the GPIO connections, the annotations are set as individual letters, "G", "P", "I", "O" , instead of "GPIO". THis causes problems when Fritzing automatically rotates this string, when the part is placed upside down (rotated 180 degree) in the schematic view. Now all letters are rotated one by one, and thus it reads "OIPG"

I will fix it for this part because I am already at it. But does this affect the other parts as well, is this intended? How was this caused?

KjellMorgenstern:
THT connectors in the PCB view only appear at once side (bottom). This would propably be a problem when producing shields for the board and must be fixed.

KjellMorgenstern:
GND is not auto detected by ground fill. Manually setting GND as Ground Fill Seed with right click works.

UniquePete:
Hi Kjell,

I have no idea why this would have happened. They certainly start out as text strings [in Illustrator], but yes, I see that this is the case in pretty much every one of the parts I’ve submitted. Not all labels are split out, only some, and not by any pattern I can recognise. In one case, two labels in a set of six, all containing three characters, are split out while the other four are not… I will try using a later version of Illustrator and see if the problem persists.

I’m afraid I don’t understand the other two problems (the one-sided THT connectors and ‘ground fill’)… but I’ve made all of my parts consistently so, if there’s a problem with one, it’ll likely be the same with all of them.

Obviously it would be best to fix these problems at the source, rather patching things up at the ‘end’, but with the exception of the text problem (and if Illustrator is exporting text to .svg files as individual characters I’m going to be a bit stuck even there) I really don’t know what needs to be done.

Pete

UniquePete:
OK, this is a problem with the version of Illustrator that I have been using. I have access to a later version, and it seems to behave as expected (i.e. it is not splitting out the characters).

I will, therefore, need to go through and redo the schematic layers of all of my parts.

Peter van Epp offered a number of other comments in the context of working with one of these parts, more than I’d care to include here, but in relation to the schematic he noted:
> schematic
> 
> ungrouped and rescaled the svg. In a text editor changed all the stroke-widths to be 10, the terminalId height and width to be 10 (making the terminalId a 10thou by 10thou rectangle) and checked the alignment (which was out slightly because the original values were smaller.) Then changed the text to match breadboard (and the Kit 32 pinout pdf.) Added text-anchor:start (right side pins) text-anchor:middle (center labels and the pin numbers on left and right) and text-anchor:end for the right side pins so changes in text length don’t change text alignment. Resize regroup (and name the group “schematic” to match the svg) and save as plain svg.
> 
Given that I’m going to have to go into Illustrator to produce new schematic .svg files, is there anything else I should to be fixing while I’m there? You mentioned the font size at one point (I can search back through the posts to find that if need be)… anything else?

Thanks
Pete

KjellMorgenstern:
The issue about the ground seed was the name in the fzp file.
Also I fixed the top layer label name. See commit cb90c3d .

KjellMorgenstern:
Typically , Fritzing schematics have the GND connector at the bottom, V connectors at the top, and other connectors grouped at the side. If you want you can adapt that, but not all boards use that style. I don't see issues with the font size, maybe that was already fixed before.

The grid alignment is also a minor thing, I think. The right side of the box runs right of the grid, while each other part uses a rectangle that runs perfectly inside the grid, so parts optically align:
image

UniquePete:
A few further queries, now that I'm back to this...

1. If I just use one ground pin at the bottom of a schematic, how to I associate several actual pins with that one graphic element? I am using the Parts Editor to set this stuff up and it only seems to let me connect a single 'physical' pin to any schematic input/output. Would I need to edit the .fzp file 'manually' to achieve this? (I'm already spending a fair bit of time directly editing the .fzp file, so this is probably no big deal if this is what is required.)

2. Once again, I am currerntly using the Parts Editor to establish the base for the various 'views', and I just import the relevant .svg files. They just get imported and placed, aligned with the grid defined therein. It's not clear to me how I would align the schematic any other way, or what I would align it with if it needed to be aligned with something else.

3. With respect to the ground seed, does this mean that all ground pins should have GND in the the 'Name' field, as well as any descriptive reference in the 'Description' field? (Again, in the Parts Editor, in the 'Connectors' section?

PR #248, opened 2020-03-09
Arduino Pro Mini 35-pin Clone (3V3/8MHz)

UniquePete:
Arduino Pro Mini 35-pin Clone, 3.3V/8MHz variant


mMerlin:
That still has the referenceFile attribute in the module tag of the fzp files.

PR #258, opened 2020-08-23
Removed black bars, which should not be there (in PCB view)

majoge:


mMerlin:
I do not think there is a issue associated with this. I saw it only in [this forum post](https://forum.fritzing.org/t/pcb-view-ic-with-package-so-w-16/10625), which was structured in the form of an issue, and tagged as a bug. I think @hector2007 thought that was actually filing a but report.

majoge:
@KjellMorgenstern Afaik, there is no github issue for this PR

PR #259, opened 2020-09-30
Added a library for Arduino MKR WiFi 1010

MEl3dwi:


mMerlin:
This is not a library, or an update to the core library either. It is a single part file and a separate part definition file. The part file I assume contains the view svg images. To be added to the core library, those need to be in the svg/core/«viewname» folders, with appropriate names. There is a naming convention for the svg files, to make them easier to reuse for additional parts. See "4. File Naming Conventions" in http://fritzing.org/fritzings-graphic-standards.

The part definition file that I can see without downloading and extracting needs some work too. The description seems to be only formatting information without any content.

PR #276, opened 2020-11-05
Obsolete core/hc-sr04_ultrafix.fzp and core/HC-SR0 Ultrasonic Distance Sensor.fzp

KjellMorgenstern:
Note: Only the obsoletion formalism is done, the part is not yet fixed.


KjellMorgenstern:
Fixed by 1348228

PR #329, opened 2021-05-12
Adding Infineon BLDC IFX007T Shield

r-nikhita:


mMerlin:
403 modified files needs a lot more description than "Adding Infineon BLDC IFX007T Shield" It would appear that some sort of bulk process was run. What was it? What was it supposed to do? Even general cleanup needs a description about WHAT the purpose is. Sliding it in as part of an "add* is not good.

pandayswarnam:
Hi @mMerlin,
This part is a motor driver shield from Infineon portfolio. This is the first time I wanted to have Infineon parts in fritzing tool. We followed the contribution guideline and did the changes as per the readme, then opened a pull request for the same. If this seems not to be the correct way, kindly suggest.

thanks 
Swarnam

mMerlin:
Your description of the submission process is correct, however that should have resulting in only 4 or 5 new files. not 400+ modified. Something you did in the fork of the repo modified a lot of existing files. It should have only added new files for the new part. One fzp part description file plus up to 4 svg file (one for each view). There should not have been any modifications to pre-existing files.

Issue #253, opened 2020-04-22
Raspberry PI3 pcb is slightly incorrect.


vanepp:
While making a Pi 4B part from the current PI3 part it appears that the GPIO connector in the current 3 part (and maybe the other PIs, I haven't checked yet) doesn't match the RPI mechanical drawing available here : 

https://www.raspberrypi.org/documentation/hardware/raspberrypi/mechanical/rpi_MECH_3b_1p2.pdf

(the center line of the GPIO header is a little high.) I corrected that for the 4B part and will eventually redo the PI3. However in case eventually never comes or someone else wants to do it, I'll document the issue here. The center line of the GPIO pins should be the same as the center line of the mounting holes, as we see here, it is not:

![Capture](https://user-images.githubusercontent.com/16514277/80042661-78917600-84b4-11ea-92fd-a90e63d77028.PNG)

There are a number of other issues with this (and the other RPI parts) that should be fixed as part of a general cleanup of core parts. 

Peter

KjellMorgenstern:
@vanepp Did you work on this? Otherwise I'd look into this next.

vanepp:
No, I haven't gotten to it yet so feel free.

Peter

Issue #203, opened 2017-10-10
Adafruit HUZZAH ESP8266 breakout board Gerber file fail


chrisgimson:
Exporting to a gerber file produces a solid copper block on the top layer underneath the component that shorts out all of the pins!.  You do not see this on the Fritzing PCB view nor on exported image files - only on the gerber file which you will be blind to unless you check the gerber files with a viewer.
The other ESP8266 parts in the library do not have this problem so the short term fix is to simply avoid the adafruit part if you are using gerber files.

KjellMorgenstern:
The part has a silkscreen component inside the copper1 component
```
<g id="copper1">
...

                    <g gorn="0.1.0.0.0.0.6"  id="silkscreen">
                            <line x1="216.135" x2="119.471"  fill="none" stroke-width="1.1114" stroke="#FFFFFF" y1="446.456" y2="446.456"/>
                            <line x1="119.471" x2="119.471"  fill="none" stroke-width="1.1114" stroke="#FFFFFF" y1="446.456" y2="500.611"/>
                            <line x1="119.471" x2="216.135"  fill="none" stroke-width="1.1114" stroke="#FFFFFF" y1="500.611" y2="500.611"/>
                            <line x1="216.135" x2="216.136"  fill="none" stroke-width="1.1114" stroke="#FFFFFF" y1="446.456" y2="459.806"/>
                            <line x1="216.135" x2="216.135"  fill="none" stroke-width="1.1114" stroke="#FFFFFF" y1="487.25" y2="500.611"/>
                    </g>
</g>
```

KjellMorgenstern:
Solved by 17fbf390930

Issue #161, opened 2018-08-20
Minor Issue in Schematic View for "SparkFun - storage microSD" part


bboyho:
Hi,

There was an issue filed in the SparkFun Fritzing Parts repo for a part that was created in the official Fritzing core library => [ [https://github.com/sparkfun/Fritzing_Parts/issues/15](https://github.com/sparkfun/Fritzing_Parts/issues/15) ]. Overall, the part is good except for the part's  "Schematic" view: the connectors for GND and VCC are not associated correctly with the .svg graphic.

<p align="middle"><a href="https://cloud.githubusercontent.com/assets/1015439/25646059/5b1e27a6-2f82-11e7-9033-d234d42feebe.png"><img src="https://cloud.githubusercontent.com/assets/1015439/25646059/5b1e27a6-2f82-11e7-9033-d234d42feebe.png" alt="error in Fritzing Schematic View, screenshot taken from user jmwright" title="error in Fritzing Schematic View, screenshot taken from user jmwright"></a></p>

I tried to find where it was "linked" in Fritzing's core library but I am not sure which file to correct in order to link the pins for the fritzing-parts repo. I exported the part as a new .fzpz here in case anyone needs => [ [https://github.com/sparkfun/Fritzing_Parts/blob/master/products/00544_SparkFun_storage_microsd_v1b.fzpz](https://github.com/sparkfun/Fritzing_Parts/blob/master/products/00544_SparkFun_storage_microsd_v1b.fzpz) ].

KjellMorgenstern:
c54d1bff

Issue #101, opened 2017-07-29
RFM12b Breadboard View Connection points wrong


GogoVux:
In Breadboard View are the connectionpoints of nIRQ and nRes exchanged (nIRQ goes to nRes and vice versa). This is only in Breadboard View.

![fritzing_rfm12b_issue](https://user-images.githubusercontent.com/25881854/28744509-d7d7f086-7462-11e7-8127-0c5b50125276.png)

PR #336, opened 2021-08-28
Milestone 0.9.9 Issue Resolve

ctice36:
Resolved Issues:
#253 
#161 
#203 
#101 
#43


KjellMorgenstern:
#203 seems solved

KjellMorgenstern:
#101 seems solved

KjellMorgenstern:
#161: I think GND and VCC terminal points where mixed up

KjellMorgenstern:
#253 seens not fixed. The file that was added to the root folder of fritzing parts is identical to the old part Raspberry Pi 3.fzp

KjellMorgenstern:
In general, parts should not be placed into the root folder.

Fixes should either be done directly in place.
Or, in cases of a breaking change,  a copy of the original part should be made in the "obsolete" folder, and the fixed version should take its place.

KjellMorgenstern:
Fixes have been applied in separate commits

PR #338, opened 2022-01-06
add new part meadow micro v2

kainazzzo:
Adding a new part for the meadow micro v2 board. Breadboard svg provided by @jorgedevs


KjellMorgenstern:
One of the automated tests failed. Without a layer id, the element might not be properly rendered in Fritzing, e.g. during gerber or svg export, or when showing in one of the views.
svg/core/breadboard/meadow_micro_v2_5.svg child element 'rect' with no layer id

PR #360, opened 2023-02-25
changed NAND gate model to use MOS transistors

failiz:
to be used in the example in the fritzing-app repository

### DIFF
--- a/core/4011_4_x_2_input_NAND_gate_multipart.fzp
+++ b/core/4011_4_x_2_input_NAND_gate_multipart.fzp
@@ -40,23 +40,19 @@ p, li { white-space: pre-wrap; }
     <line>x{instanceTitle}B {net connector4} {net connector5} {net connector3} {net connector13} {net connector6} NAND</line>
     <line>x{instanceTitle}C {net connector7} {net connector8} {net connector9} {net connector13} {net connector6} NAND</line>
     <line>x{instanceTitle}D {net connector11} {net connector12} {net connector10} {net connector13} {net connector6} NAND</line>
-    <model>.SUBCKT NAND 1 2 3 4 11
-    * NODES : INPUT (2) , OUTPUT , VCC , GND
-    Q1 9 5 1 QMOD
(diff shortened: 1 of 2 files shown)


failiz:
Hi Kjell,
the NAND gate would be very interesting to be used in examples in the transitory branch. Currently, the NAND gate that is in the core uses BJT transistors and cannot drive LEDs (and the sketch that I made does not work well). The NAND gate with multiple subparts has a problem that you identified and it should not be merged yet. 
Is it Ok if I add the changes to the spice model of the NAND gate (single part) in the ac branch and close this PR?

KjellMorgenstern:
Maybe we can add a buffered/driver NAND as additional part? With a quick search I only found this
CD40107BE , which is only 2 Gate, but still THT.

And the logic level NAND gate would then be a good example to build the same, using additional transistors to drive the LEDs?

failiz:
Context: I did not model any NAND gate, I just copied the models from the ngspice manual. And there are two of them, the one based on BJT transistors and the one using MOSFETs. I made the NAND example (https://github.com/fritzing/fritzing-app/pull/4004) and I realized the BJT NAND model was giving problems, so I changed to the MOSFET version. 
We should not create two parts, one more "logic level" and another able to drive/sink more current. Any IC with NAND gates will have enough capability to drive an LED. Probably, the BJT NAND gate could also be modified to drive LEDs directly by tuning the model of the BJT transistor used (but I am nor an expert in SPICE modeling and it could take a while and most gates are based on MOSFT technology nowadays). So, I just think we should merge the new version of the single NAND part (not the multipart NAND gate) and the NAND example. It is a nice example and I could use the NAND gate to create interesting examples for the transitory branch.

failiz:
@KjellMorgenstern , could you merge this PR? It will make my life easier and I think this spice code should be faster (4 transistors versus 5 transistors and 3 diodes of the previous spice) and more representative of the current technologies (I doubt that the gates now are using BJT transistors)

Issue #347, opened 2022-09-11
obsolete.py improvements


vanepp:
I'm learning and documenting obsoleting, and as part of that have been poking at obsolete.py. Before I get too far I figured I should explain what I want to change and see if there is agreement or if I am overlooking something. 

bug:

The dom returned by get_dom isn't the root, but rather     return dom.documentElement and thus 

 s = new_fzp_dom.toxml("UTF-8", False)

doesn't emit the  <?xml version='1.0' encoding='UTF-8'?> line (because the dom isn't the root) on the output fzp files (obsolete and core.) The above format also requires python 3.9 or later to add the standalone=no parameter (which most Fritzing fzps have.) Returning dom from get_dom and adding the .documentElement to all other dom references fixes that. I don't think that Fritzing cares, because UTF-8 is the default, but vim doesn't recognize the file as xml (and thus doesn't enable the xml checks.) I expect this shouldn't be controversial. 

Next is an enhancement which may be more controversial. I'd like to remove the revision argument and replace it with the version in the fzp file with the new fzp and the svgs getting the current version plus 1. There are at least a couple of files in core parts that have a version 1.1 currently, so they would need to be updated to be 1 (so the version is an int.)  If there is no version tag in the fzp file the code currently sets it to 2, and I think that 1 is a better choice (with the new fzp being 2.) In the case of a fzp with a version number the new version would be the current version + 1. I'm not sure if the revision argument is intended to allow something that I'm not aware of. I think (but haven't tested yet) that the current code will break if there is a version number in the fzp file. I think this code is going to break if there is a version tag in the fzp. 

    versions = obsolete_fzp_dom.getElementsByTagName("version")
    if not versions:
        doc = obsolete_fzp_dom.documentElement.ownerDocument
        version = doc.createElement("version")
        version.appendChild(doc.createTextNode("1"))
        before_node = obsolete_fzp_dom.documentElement.childNodes[1]
        obsolete_fzp_dom.documentElement.insertBefore(version, before_node)
        print("Set version to 1 as it didn\'t exist")
    else:
        version = versions[0]
        print("Set version to %s" % version)

    new_fzp_dom = deepcopy(obsolete_fzp_dom)

    version.setAttribute("replacedby", new_module_id)

[...]

PR #364, opened 2023-03-06
obsolete.py script improvements

vanepp:
An replacement of the original pull request described in issue #347 Added writing the path to bins/core.fzb and made the changes in the repository (rather than copying a file across) so the change now diffs in git

### DIFF
--- a/scripts/obsolete.py
+++ b/scripts/obsolete.py
@@ -1,6 +1,6 @@
 # Part obsoletion script
-# TODO: Support obsoleting a part without having a replacement. This could be useful for parts whic are not available anymore.
-#
+# TODO: Support obsoleting a part without having a replacement. This could
+# be useful for parts which are not available anymore.
 
 import argparse
 import textwrap
@@ -24,12 +24,12 @@ def get_dom(filename):
         print(str(err), filename)
         sys.exit(-1)
 
-    return dom.documentElement
+    return dom
 
 
 def set_module_id(dom, name):
     newModuleID = '%.8s%s' % (re.sub(r'\s+|_', '', name), uuid.uuid4().hex)
-    dom.setAttribute("moduleId", newModuleID)
+    dom.documentElement.setAttribute("moduleId", newModuleID)
     return newModuleID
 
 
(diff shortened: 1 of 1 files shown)


vanepp:
I just came across a potential issue with (probably both) obsolete.py implementations. Minidom appears sometime reformat the xml on output as in this case in core.fzb

```$ git diff develop Dual_VNH2SP30_Motor_Driver_2
diff --git a/bins/core.fzb b/bins/core.fzb
index 39706e9..145199f 100644
--- a/bins/core.fzb
+++ b/bins/core.fzb
@@ -1,11 +1,10 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<module fritzingVersion="0.4.3b.0.0" icon="Core.png">
+<?xml version="1.0" encoding="UTF-8" standalone="no"?><module fritzingVersion="0.4.3b.0.0" icon="Core.png">
        <title>Core Parts</title>
     <instances>
         <instance moduleIdRef="__spacer__" modelIndex="3" path="Basic">
             <views>
               <iconView layer="icon">
-                <geometry z="-1" x="-1" y="-1"></geometry>
+                <geometry z="-1" x="-1" y="-1"/>
               </iconView>
             </views>
         </instance>
```

The first line shouldn't be a problem, it just added the  "standalone="no"?" that I told it to. I don't think the next one where 

```</geometry>```

 gets replaced by

 ```/>```

 should be a problem either because they are identical xml (I think at least!) but I thought I should point it out here in case there is an issue I'm not aware of. In this case, this should only occur on the first instance (which will be this one when I get the pull request together!) in to core parts, after that the core.fzb file will be in the format the Minidom outputs. I expect similar issues may show up in other xml as it is obsoleted by obsolete,py but I'm not sure how we could fix it easily as it doesn't appear to be prettyprinting the xml which was my first thought. I don't see any option to pass the xml through unchanged, but I'm not all the familiar with Minidom either. I think I may be able to convince lxml to pass the xml through unmodified (although I'm not sure of that either!) but that would require loading lxml in to python which is probably less desirable than using Minidom which is built in.

PR #396, opened 2024-07-01
Calliope Mini 3

HaraldRau:
## Bauteil nach Revision und Fehler beim Importieren ##

- fzpz-Datei aus dem Forum extrahiert und umbenannt
- breadboard-svg im Zweig gelöscht und ersetzt durch neue Breadbord-svg

### Quelle: ###
[https://forum.fritzing.org/t/calliope-3-als-bauteil-nach-fritzing/24107/23](https://forum.fritzing.org/t/calliope-3-als-bauteil-nach-fritzing/24107/23)

[Calliope_mini_3-connectors-fixed.fzpz (50.3 KB)](https://forum.fritzing.org/uploads/short-url/hvyU95Uwe3pwpHjxKMD53yvRrO2.fzpz)

[Calliope_mini_3_mit_LED](https://forum.fritzing.org/uploads/short-url/4sOvf6qEVpfyMSW0UAyksJAkXio.fzpz)


HaraldRau:
Good morning,

Calliope Mini 3 from the Forum

[https://forum.fritzing.org/t/calliope-3-als-bauteil-nach-fritzing/24107/34](https://forum.fritzing.org/t/calliope-3-als-bauteil-nach-fritzing/24107/34)

Now it worked. Thanks to Peter's work, no further errors occurred.

Thank you all for your help.

KjellMorgenstern:
Connectors on the left Jacdac are labeled jadac1-*
Connectors on the right Jacdac are a mix of jadac1-* and jadac2-*

Is this intended, or maybe label them jacdac1-* and jacdac2-* ?

HaraldRau:
Yes, I think that's a mistake. The correct name should be:
jadac1-1/D
jadac1-2/GND
jadac1-3/3V3

jadac2-1/D
jadac2-2/GND
jadac3-3/3V3

Thank you for troubleshooting!

HaraldRau:
I will change it in the fzp-file.

HaraldRau:
I found some disconnected connections. They weren't labeled on the board. I2C and UART are routed to the PIN header.
I checked it with the voltmeter.
 I'll change that tonight!

HaraldRau:
I found another error. On the breadboard, pin 23 (GND) is connected to JP1. In the PCB view, this connection is not routed and shows a connection from JP1 to pin 19. I am looking for the error!
BME_280_Calliope_Steckplatine
BME_280_Calliope_Leiterplatte

HaraldRau:
Is it better to close the pull request for now?

KjellMorgenstern:
No need to close it, just push fixes to it as they come in. 

Besides, I am not sure if this is an error. Pin 19, 22 and 23 all are GND. 
Fritzing will always select the closest connector of the subnet to draw the ratsnest line.

HaraldRau:
Ah, that's right. On the old Calliope, pin 19 was I2C; now it is GND. I will still check it again!
Thanks!

HaraldRau:
I just panicked a bit; everything is fine!
BME_280_Calliope_Leiterplatte
BME_280_Calliope_Steckplatine

HaraldRau:
I think the XML code is not correct. In the previous version, the "style" element was defined at the beginning of the document. The current version has a "style" element for each object again.

Are there any invalid entries in the "style" elements?

` style="fill:none;fill-opacity:0.712946;stroke:#faf1a9;stroke-width:0.773404;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1;paint-order:stroke markers fill" `

KjellMorgenstern:
When exporting (from Inkscape) could you try "optimized", and check the SVG 1.1 box?

This will apply "scour" to optimize the SVG. It avoid conflicts between styles and attributes, possibly prevent the SVG from containing styles that Fritzing does not support. 

image

HaraldRau:
I did that, the source code looks much cleaner to me now. Thank you very much for the tip. I'll take another look at the file in Git and review the changes.
Gruß Harald!

vanepp:
@KjellMorgenstern  Where are the settings you referred to in Inkscape? I can't find them in preferences (on Inkscape 1.3.2) are they somewhere else?

HaraldRau:
The settings must be activated when saving. If you select the "Optimized SVG" settings, an optimization script is started. You can then select a few options in a dialog.
**File/save Dialog**
Bildschirmfoto vom 2024-09-04 17-59-57

**Optimierungsdialog**
Bildschirmfoto vom 2024-09-04 18-00-35

vanepp:
Ah! That works a little better, but I still don't see the dialog that allows selecting export as svg 1.1 anywhere. I have been using plain svg (which was the original recommendation from years ago) like this 

capture

changing to optimized svg does indeed bring up more dialogs but not the ones I want

capture1

capture2

capture3

nothing that sets the 1.1 output.

HaraldRau:
I'm using Inkscape 1.2.2 on Ubuntu 24.04.1. I think that's the reason for the missing option.

KjellMorgenstern:
When selecting "Remove unused IDs" , I find it useful to Preserve IDs starting with
"conn,bread,copper,silk,pcb,schematic,icon"

That list might not be complete. Not all IDs important to Fritzing end with digits.

mMerlin:
I have only "conn" in Preserve IDs starting with. For Preserve the following IDs, I have "icon,breadboard,schematic,copper0,copper1". Tweaked over time.

Scour is a python app that can be run standalone. See https://github.com/scour-project/scour. Apache-2.0 license. I'm pretty sure that anything that can selected from the Inkscape dialog can be done from the command line (or a script) as well. A possible tool to add to the CI.

@vanepp The svg 1.1 setting is (for me, Inkscape 1.3.2) on the Save As dialog. You can see it bottom left of the first image in https://github.com/fritzing/fritzing-parts/pull/396#issuecomment-2329458670. Even if it is not in English, the "SVG 1.1" text by the checkbox is enough to point to it. AFAIR, that checkbox has been there back several versions of Inkscape.

KjellMorgenstern:
@mMerlin Yes, I apply it all the time. The CI can't modify the files, but maybe this is possible: https://github.com/fritzing/fritzing-parts/issues/397

vanepp:
@mMerlin  Apparently only on Linux. On Win10 I get only this:

capture

no other options that I can find. Clicking on save brings up another menu but it also doesn't have anything to do with the svg level. I'll boot up Mint and load Inkscape and try it there and/or look over the Inkscape forums.

KjellMorgenstern:
I tried the "Optimized SVG" on Windows 10, Inkscape 1.3.2. 
The dialog mostly looks like in your screenshots above, there is no option to directly use SVG 1.1.

The screenshot I posted was from Inkscape 1.1.2 that ships with ubuntu 22.

PR #400, opened 2024-09-30
Ifx cy8 ckit 062 s2 ai

janaifx:
@KjellMorgenstern i have added one more board, kindly make it visible in Fritzing tool


KjellMorgenstern:
Looks good, I will add it to the 1.0.4 release.

The breadboard is more than 400kB SVG, which makes it 8th larges SVG in Fritzing. It can be optimized by removing a number of non visible elements, and applying scour, to about 65%, without loss or visible changes.

It works fine on my PC, but on less powerful PCs it might slow down Fritzing. However, I think it might be better to introduce these kind of optimizations for all parts, systematically.

If you observe performance problems handling the part, please let me know.

janaifx:
@KjellMorgenstern this board Infineon_CY8CKIT-062S2-AI is still not visible in Fritzing tool.
kindly please have a look...

PR #414, opened 2025-04-03
Fixed FC-51 Part

RAPTOR7762:


KjellMorgenstern:
One of the checks failed:
https://github.com/fritzing/fritzing-parts/actions/runs/14241104003/job/39912391275?pr=414

I think the old svg files should be moved to the obsolete folder for svg files, too.

RAPTOR7762:
I hope it works now.

@KjellMorgenstern I've moved the old SVGs to obsolete.

RAPTOR7762:
Weird, I had uploaded all the relevant files. Am I missing something? 

@KjellMorgenstern can you help me add the files to this REPO?

KjellMorgenstern:
The files you added, for example "svg.pcb.prefix0000_4fb85df3c2742642b3a26649700a513a_2_pcb.svg" is not the filename expected, read the error message carefully.

The trick is: "/" needs to be escaped when used in xml. To work around this (maybe it was thought a simplification, but I think it wasn't) Fritzing allows a "." . And to further complicate it, Fritzing has part categories like "core" and "contrib" and "obsolete" which are magically added to the path.

Therefor, a filename "svg.pcb.name.svg" can be read as "svg/core/pcb/name.svg".

PR #419, opened 2025-06-19
Infineon_CY8CPROTO-062-4343W

KannanInfn:


KjellMorgenstern:
There are several errors detected with this part, most of the concerning the connectors.

There are conflicting style and stroke attributes. The style attribute should be removed, to avoid ambiguity.

Many of the connectors are not visible. This might be intended or not, but is very error prone (you can't see them, so very hard to detect leftovers or wrongly positioned connectors). There only a tiny amount of use cases where you really want connectors to not be visible, so this should be fixed.

Also there are some connectors referenced that don't exist. This is usually miss-typed connector name, or copy and paste error when using a different part file as starting point (forgetting to remove connectors).

https://github.com/fritzing/fritzing-parts/actions/runs/15775773430/job/44469769661

KjellMorgenstern:
The error from scripts/connectors_misnumbered.py is unrelated. 
The script itself is going to be replaced with the scripts/checks
check-all-parts is unrelated, and the script will be replaced with the scripts/checks/ fzp_checker_runner.py , which already passes.

So this is fine, if there are any more issues I'l let you know.

Issue #409, opened 2025-03-20
fc-51 parts in core parts have a problem in pcb.


vanepp:
The two fc-51 parts (simulation off and simulation on) have pcb pins reversed. See forum post

https://forum.fritzing.org/t/parts-are-mirrored-in-pcb/27736

Here are two parts that correct the problem, but I expect there is going to be an issue with obsoleting the existing parts (although hopefully I am wrong!) Since the pcb pin position has changed without the pin number changing I think the replacement part will route incorrectly in an existing sketch and I don't see a way around it, as the connector Id isn't changing in the fzp file. Hopefully you do see a way around it. I also took the opportunity to update the parts to meet the graphics standards. The two parts are here (as well as in the forum post.) Remove the trailing .zip to get the .fzpz file. 

[fc-51-IR-Sensor.fzpz.zip](https://github.com/user-attachments/files/19376845/fc-51-IR-Sensor.fzpz.zip)

[fc-51-IR-Sensor_On.fzpz.zip](https://github.com/user-attachments/files/19376859/fc-51-IR-Sensor_On.fzpz.zip)

KjellMorgenstern:
We could
- add the new fc-51 sensor parts
- move reverse orientated fc-51 to obsolete
- *not* set a replasedby id on the obsolete part, so existing circuits will just keep using it. We assume these circuits already have been validated with the existing VCC - GND - OUT order
- unlist the old and list the new parts in the core and the sim bins.

vanepp:
That sounds like a good solution if you are agreeable to it. The issue I see is that parts that are routed in pcb in a sketch will suddenly break as the routing will change (pins 3 2 1 will become 1 2 3 and the traces will likely cross.) Unfortunately I don't see a way around that though. Ah! I think I see, the old sketch will still work because the old part is in obsolete and will be used (but be in the wrong order but still work as it did) and because it isn't obsoleted you won't be prompted to update it.  Yes that seems to be a very good solution that won't break anything unexpected as far as I can see. Good thinking (I was hoping you would be able to think of something I hadn't thought of and you did!)

RAPTOR7762:
Yes, I agree with Peter's suggestion.

@KjellMorgenstern shall we go ahead with obsoleting the part (or will cause anything catastrophic?)

KjellMorgenstern:
This way we would just add a new part, and hide the one with the irritating connectors. There is not much that can go wrong.

[...]

Issue #369, opened 2023-03-24
Parts policy discussion


vanepp:
This is a continuation (with a better topic name) from #367 . The intent is to modify and possibly write new standards for parts. We have the part file format document

https://github.com/fritzing/fritzing-app/wiki/2.1-Part-file-format

and the graphics standards

https://fritzing.org/fritzings-graphic-standards

both of which may need updates as part of this discussion, but which provide the basis for this discussion. 

@KjellMorgenstern @mMerlin  @failiz where suggested in #367 as people who need to be involved, but I expect anyone else that is interested should also be welcome to join in. 

 My broad outline is to get a parts policy in place that allows automated checking of parts and  consistency  of parts (at least those included in core parts) so schematic as one instance looks the same and interoperates well with other parts. The longer term goal is, once a policy is in place, to start a project to update the parts in core parts to meet the new standard. (that is likely to be a lot of work, but requires the policy to be in place to allow it!)  That would also allow the automated checking of parts that are submitted to core parts to make sure they meet the standards. However it leaves the user able to create custom parts that don't meet the standards (either due to time constraints or lack of need of some views) for their personal use which has been one objection raised in past discussions on this topic. Another topic is the layout of the repository as discussed in  #4018 . Documenting the format of the .fzpz file and its association / conversion to the repository format should be done. I am not aware of any documentation on the format of the .fzpz file existing (pointers to anything I have missed are welcome!)  Documentation on each field in the fzp file would be useful. For instance the version field is often set to 4 where I think it should start at 1 for a new part and increase by one for each new version of the part. My update to the obsolete.py script available in #347 addresses that by setting the version field to 1 if there isn't one already or increasing the current version field by 1 if there is one.. Issue #365 to standardize the parts labels (which affects the label field in the fzp file) also enters in to this discussion as I think it is desirable to standardize the label field in parts. File name conventions for fzp and svg files are also an issue that needs to come up. I would prefer that the version number appear in the files so for the first version of the part the fzp and svg files end in _1 such as part.fc-51_1.fzp for the first version of a part. One question would be is there a need to add a guid field (a globally unique identifier) to the file name as parts editor currently does when editing parts? As noted I think we need to update the part file format document to specify all the fields in a fzp file along with their suggested initial value and what they relate to in other files. As an example 

```
 <breadboardView>
      <layers image="breadboard/fc-51_breadboard.svg">
        <layer layerId="breadboard"/>
      </layers>
  </breadboardView>
```

[...]

PR #420, opened 2025-06-23
FC-51 3rd Attempt

RAPTOR7762:
Work in progress


RAPTOR7762:
Hi Kjell, once I'm done can you help me run this against it, befor I mark this as completed; ready for review?

KjellMorgenstern:
The current behavior of Fritzing, to add a prefix0000 is a bit confusing, and only very limited help to avoid duplicate filenames.  
Using a uuid in the filename works to avoid name collisions. In the part format discussion, we currently drafted
in https://github.com/fritzing/fritzing-parts/issues/369 :
"File name conventions: Use a "_versionnumber" or a "_hash" at the end of the file name. e,g: fc-51_1.fzp. Avoid spaces and brackets. Use underscore symbols."

Although this was for the fzp filename, it could be applied to the views:
So, fc-51_3.svg would match this, but you could also use fc-51_0c9711422ad29606f2b6bb1ec42bc763.svg .

It might be more descriptive to put the _[view] in there, but that wil be duplicate information in many cases:
The view is defined in the fzp file itself. Again, for fzpz, with a [view]. prefix. And for fzp files by the svg/[view]/ directory name. And also, within the svg, with the layer id.
Also, when looking at the actual image, it is usually very obvious which view it is, except maybe for icon views.

RAPTOR7762:
i think there are minor errors and one major one -- the trailing px for font size in icon SVG

RAPTOR7762:
Hi @KjellMorgenstern Are you going to merge this and add to Fritzing 1.0.6? After looking at the logs there isn't anything wrong with the part

KjellMorgenstern:
In the description of a PR, you should link back to the issue, so we don't need to guess what the original problem was.
Something like:
Fixes #409

RAPTOR7762:
Ok will do that in the future!

RAPTOR7762:
OK @KjellMorgenstern I've added the `changedby=""` attribute

RAPTOR7762:
Hi @KjellMorgenstern 

Are you going to add this to v1.0.7

PR #422, opened 2025-07-11
Adafruit Bin

RAPTOR7762:
Hi @KjellMorgenstern 

Added the very first part -- MCP3008 (updated). Can you run the workflow?


RAPTOR7762:
Adding one more part

RAPTOR7762:
Adding no-header version

RAPTOR7762:
Add TC74 IC (TO-220)

KjellMorgenstern:
Were the terminalId's broken or ophaned? Otherwise, we should just keep them to keep the parts as original as possible.

I guess setting reference to to original module ID will be helpful, so the part can keep a reference to the original fzpz from Adafruit.

We will need new module IDs, other wise the IDs would conflict with users and sketches that already have a fzpz from Adafruit loaded.

Adding these parts one by one seems not very feasible to me. I think a python script should be used for that.

How many parts are you planning to add? What is your timeline to get this done?

RAPTOR7762:
Hi @KjellMorgenstern 

>Were the terminalId's broken or ophaned? Otherwise, we should just keep them to keep the parts as original as possible.

Originally, I downloaded the parts made by `@BlitzCityDIY` to edit as there were errors. Being lazy, I redid everything and used the generic IC as a base. There will then be stray terminalIds in the `.fzp` as the `.svg` does not. To solve, I removed the terminalId definition in the `.fzp` file

>I guess setting reference to to original module ID will be helpful, so the part can keep a reference to the original fzpz from Adafruit.

OK. I was thinking of that too, but I thought it would be easier to trace the file and make addition of parts into bins easier

Therefore, the new IDs as you mentioned:

>We will need new module IDs, other wise the IDs would conflict with users and sketches that already have a fzpz from Adafruit loaded.

>How many parts are you planning to add? What is your timeline to get this done?

So I'll keep to adding 1-2 per day. Assuming that Iadd 2/day, then:

* The first wave will have ~160 parts, so
* 160 / 2 = 80 (days)

Then I'll add a few more waves (depending on when you're going to release 1.1.0

>Adding these parts one by one seems not very feasible to me. I think a python script should be used for that.

OK. I'll run through @vanepp's FritzingCheckPart.py and your checker runners (I just need to figure out yours) and see if I can get an answer

KjellMorgenstern:
I was more thinking of a custom python script, to run on all the ~1000? parts and migrate them.

From a database perspective, this really is a migration. 

Writing the script would be about 1-2 days.

I recently added fzpz support for the check script, running it on the adafruit repo could give an estimate about how many days will be needed for cleanup.

RAPTOR7762:
Tganks a lot!

RAPTOR7762:
***

Adding the Adafruit Pi Stemma QT Breakout

RAPTOR7762:
Add Adafruit Raspberry Pi Bin

RAPTOR7762:
Adding PIR sensor (no bendable legs)

RAPTOR7762:
Adding PIR sensor (bendable legs)

RAPTOR7762:
Add Adafruit 2.13" 250x122 Quad-Color eInk / ePaper Display w/ SRAM

RAPTOR7762:
Back to MCP3008: forgot to upload SVGs

RAPTOR7762:
Add ItsyBitsy bin

RAPTOR7762:
Add Adafruit QMC5883P - Triple Axis Magnetometer - STEMMA QT

RAPTOR7762:
Sync with bin

Issue #426, opened 2025-10-20
Add a solar panel


KjellMorgenstern:
A good start could be the part here:

https://forum.fritzing.org/t/thesis-project-opinion-cooling-pv-panel-methods/28474

It would be nice to have the panel configurable, but not a requirement.

RAPTOR7762:
Hi @KjellMorgenstern What do you mean by "configurable"

PR #427, opened 2025-10-25
Add generic small 6V solar panel

RAPTOR7762:
@KjellMorgenstern Addressed #426. Can you add this part to Fritzing 1.0.7? Ready for review


KjellMorgenstern:
The part is scheduled for 1.0.7.

I see I need to update the github workflow to run the checker.

KjellMorgenstern:
@RAPTOR7762 There is a report about some (minor) problems here:

https://github.com/fritzing/fritzing-parts/actions/runs/18935939327/job/54062887256#step:7:92

The latest version of the checker script should be able fix these automatically if you run it with the "--fix" option.

Of course you can also fix manually, for some errors there is not automatic fix. But hopefully the new report after running the check is now much better to read.

KjellMorgenstern:
A markdown version of the result is also shown here: https://github.com/fritzing/fritzing-parts/actions/runs/18936649301?pr=427 .

RAPTOR7762:
OK fixed @KjellMorgenstern

RAPTOR7762:
@KjellMorgenstern ?

failiz:
Nice part. However, the schematic symbol could be improved by (1) shortening the long stroke to not touch the outside circle and (2) follow the convention in electronics where positive is up and negative down or, in other words, rotating the symbol 90 degrees so the positive side is up and the negative down. I did not test this in Fritzing and maybe it already appears like that when you drag it to the program, but the svg appears rotated.
The second issue is present in several batteries in the core, and it is a bit difficult to fix without obsoleting them. But this part is new and easier to fix...

PR #429, opened 2026-03-19
Infineon_BLDC-SHIELD_IFX007T

KannanInfn:


KannanInfn:
@KjellMorgenstern 
I have added one more board. Kindly check the errors and provide your feedback.

KjellMorgenstern:
In the breadboard and icon file: Nesting breadboard and icon groups isn't supported and could have side effects now or in the future. 

In the fzp: The fritzingVerison is set to some ancient version. This part wasn't possibly tested with that version of Fritzing. I think you should set it to 1.0.6.

KannanInfn:
> In the breadboard and icon file: Nesting breadboard and icon groups isn't supported and could have side effects now or in the future.
> 
> In the fzp: The fritzingVerison is set to some ancient version. This part wasn't possibly tested with that version of Fritzing. I think you should set it to 1.0.6.

@KjellMorgenstern Thanks for the feedback.
The files are updated and ready for review. Kindly do the needful.

